PR #64, opened 2017-09-28
add types for descriptor and manifest

mengqiy:
We finally agreed to land on kubectl repo.
Moved from main repo PR: https://github.com/kubernetes/kubernetes/pull/52570.
A lot of discussion there.

### DIFF
--- a/pkg/apis/manifest/v1alpha1/doc.go
+++ b/pkg/apis/manifest/v1alpha1/doc.go
@@ -1,0 +1,19 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


mengqiy:
cc: @bgrant0607-cc PTAL

bgrant0607:
Thanks. 

I'm currently thinking we should break out Descriptor into a CRD definition somewhere (possibly not in this repo) and put a reference in Manifest to an "owner". Kubectl could then set ownerReferences to the owner automatically. The owner could be the Descriptor CRD or another object. Kubectl doesn't actually need the Descriptor for anything during expansion.

But further changes can be made in subsequent PRs. Merge and iterate.

/lgtm

bgrant0607:
A followup comment:

The core parts of the manifest proposal are simply declarative expressions of existing kubectl commands and flags: kubectl create configmap, kubectl create secret, kubectl patch, kubectl label, kubectl annotate, --filename, --prune, --recursive, --selector.

PR #65, opened 2017-09-28
start kexpand

monopole:
A place to hang code.

Starting with `kexpand` since `expand` is already taken on linux systems, and since it's
unique enough to change with `find ./ | xargs sed -i 's/kexpand/foo/'`

### DIFF
--- a/cmd/kexpand/README.md
+++ b/cmd/kexpand/README.md
@@ -1,0 +1,42 @@
+# kexpand
+
+_TODO: flesh out this placeholder documentation_
+ 
+`kexpand` is a kubernetes cluster configuration utility,
+a prototype of ideas discussed in [this doc][DAM].
+
+It accepts one or more file system path arguments,
+reads the content thereof, and emits kubernetes
+[resource] yaml to stdout, suitable for piping 
+into [kubectl] for [declarative] application to a
+kubernetes cluster.
+
+For example, if your current working directory contained
+> ```
+> mycluster/
+>   Kube-Manifest.yaml
+>   deployment.yaml
+>   service.yaml
+>   instances/
+>     testing/
+>       Kube-Manifest.yaml
+>       deployment.yaml
+>     prod/
+>       Kube-Manifest.yaml
+>       deployment.yaml
+>  ...
+> ```
+
+then the command 
+```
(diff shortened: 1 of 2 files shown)


cmd/kexpand/README.md
```diff
@@ -0,0 +1,42 @@
...
+It accepts one or more file system path arguments,
+reads the content thereof, and emits kubernetes
+[resource] yaml to stdout, suitable for piping 
+into [kubectl] for [declarative] application to a
+kubernetes cluster.
+
+For example, if your current working directory contained
+> ```
+> mycluster/
+>   Kube-Manifest.xml
```
bgrant0607: You don't actually mean XML, I hope. :-)

cmd/kexpand/README.md
```diff
@@ -0,0 +1,42 @@
...
+>       deployment.xml
+>     prod/
+>       Kube-Manifest.xml
+>       deployment.xml
+>  ...
+> ```
+
+then the command 
+```
+kexpand ./mycluster | kubectl apply -f -
```
bgrant0607: I disagree it should work this way. I'd expect:
```
kexpand -f ./mycluster/instances | kubectl apply -f -
```

1. Use the same builder/filename pattern as kubectl bulk commands
2. Instances should be the direct targets rather than the bases

I suppose `kexpand -f ./mycluster --recursive` should work even if the bases were discovered first, but it seems likely that users will shoot themselves in the foot if we advocate that usage.

monopole: Changed to instances.  In my head i had --recursive turned on by default.

Perhaps we should target one instance, but this is just to get started anyway.

As for `-f`, i omitted it since this program must have at least one file arg (like `cp`, `mv` etc)  and since it may one day operate as a kubectl plugin (flag collisions may be a bit dicey at the moment).  Can use kubectl builder behavior regardless.

cmd/kexpand/README.md
monopole: lol, sorry fixed

bgrant0607:
/lgtm

If we make this a kubectl extension, we'll want to follow kubectl flag conventions eventually.
https://github.com/kubernetes/community/blob/master/contributors/devel/kubectl-conventions.md#flag-conventions

PR #102, opened 2017-10-23
add kexpand util

mengqiy:
Move some util code from https://github.com/kubernetes/kubernetes/pull/52682 to kubectl repo.

There will a followup PR to move https://github.com/kubernetes/kubernetes/pull/52682/files#diff-0c3f4528c27638ae0b2b2fd6e2ecff6b code to kubectl repo. But it is blocked by some dependency issue of kubectl factory code.

### DIFF
--- a/pkg/kexpand/util.go
+++ b/pkg/kexpand/util.go
@@ -1,0 +1,121 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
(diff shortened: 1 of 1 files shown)


mengqiy:
It seems we don't have merge automation for kubectl repo. Do we want to set it up? cc @apelisse

PR #103, opened 2017-10-24
Add travis CICD coverage.

monopole:
Also, rename "kexpand" to "kinflate" to ease pronunciation.


pre-commit.sh
```diff
@@ -0,0 +1,15 @@
...
+#   cat /dev/null | read
+# exits with $? == 1, and the
+# leading `!` negates that to success.
+
+! gofmt -s -d -l . 2>&1 | read
+! goimports -l . | read
+! go vet -all ./... 2>&1 | read
+! golint ./... | read
+
+go test k8s.io/kubectl/cmd/kinflate/...
```
mengqiy: I think it not necessary to have this line, I believe your travis config will auto run all go tests.

monopole: I wanted to have a pre-commit script i could run manually.
My read of https://docs.travis-ci.com/user/languages/go/#Default-Build-Script
suggests that specifying "script" overrides the default behavior.

PR #129, opened 2017-11-13
Use deps

apelisse:
Use github.com/golang/deps to manage dependencies. It's cleaner.


apelisse:
It's difficult to know what is used and what is not in that repo, since it's doing nothing for the moment. But I removed the code (unused?) that depends on resources from k8s.io/kubernetes for now. I don't think it'd be a good idea to vendor kubernetes in kubectl.

Gopkg.toml
```diff
@@ -0,0 +1,30 @@
...
+#  name = "github.com/x/y"
+#  version = "2.4.0"
+
+
+[[constraint]]
+  branch = "v2"
+  name = "gopkg.in/yaml.v2"
+
+[[constraint]]
+  branch = "master"
```
mengqiy: I don't think pin to master branch is good idea.

pkg/kinflate/util.go
```diff
@@ -17,63 +17,17 @@ limitations under the License.
...
 )
 
 const kubeManifestFileName = "Kube-manifest.yaml"
 
-// loadBaseAndOverlayPkg returns:
-// - List of FilenameOptions, each FilenameOptions contains all the files and whether recursive for each base defined in overlay kube-manifest.yaml.
-// - Fileoptions for overlay.
-// - Package object for overlay.
-// - A potential error.
-func loadBaseAndOverlayPkg(f string) ([]resource.FilenameOptions, resource.FilenameOptions, *manifest.Manifest, error) {
```
mengqiy: Don't remove it.

Comment it for now :)

apelisse: Yeah, I'll comment ;-)

Gopkg.toml
apelisse: This is what `deps`has set by default. what do you suggest?

mengqiy: If production kubectl code lives here, I will suggest use `version = "v1.8.3"` to make it more stable.
But no production code lives here now ATM, so I take it back. Using the `master` is fine.
Keep it `master` until kubectl is isolated and moves here.

apelisse:
Tests are failing on this PR since the "verify travis" script doesn't work well with the vendor directory. We need to merge #130 first.

pkg/kinflate/util.go
apelisse: Done

apelisse:
@mengqiy I would really appreciate if you could review and maybe approve this change.
@hoegaarden Can you please review this quickly, even if its merged when you have time to do it?

Thank you!

mengqiy:
I have to say using dep to manage `apimachinery` repo will make `kinflate` work harder.

mengqiy:
See: https://github.com/kubernetes/kubectl/pull/144/commits/68011dc6ca9393c263ce924aee714c55b7f646bc#diff-a6c2ab3f6dea554a6f402bf4a9f269cfR73

PR #130, opened 2017-11-13
Simplify travis script, work against `vendor/`

apelisse:
The current travis script currently fails if we have a `vendor/`
directory, and is also quite longer than necessary.

### DIFF
--- a/bin/pre-commit.sh
+++ b/bin/pre-commit.sh
@@ -1,46 +1,25 @@
 #!/bin/bash
-# Runs pre-commit tests.
-#
-# Instead of failing on first error, complete all checks, then fail if need be.
 
-# Echo go's version to verify against go's behavior below.
-# Different versions do different things with respect to vendored code.
-go version
+rc=0
 
-# Assert state.
-if [ -n "$failIt" ]; then
-  echo "Expecting failIt to be empty."
-  exit 1
-fi
+go_dirs() {
+  go list -f '{{.Dir}}' ./... | tr '\n' '\0'
+}
 
-wantEmpty=$(gofmt -s -d -l . 2>&1 )
-if [ -n "$wantEmpty" ]; then
-  printf >&2 '\ngofmt failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running go fmt"
+go_dirs | xargs -0 gofmt -s -d -l
+rc=$((rc || $?))
 
-wantEmpty=$(goimports -l $(find . -type f -name '*.go' -not -path "./vendor/*") 2>&1)
-if [ -n "$wantEmpty" ]; then
-  printf >&2 '\ngoimports failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running goimports"
+diff -u <(echo -n) <(go_dirs | xargs -0 goimports -l)
+rc=$((rc || $?))
 
-wantEmpty=$(go vet -all ./... 2>&1)
-if [ -n "$wantEmpty" ]; then
-  printf >&2 '\ngo vet failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running go vet"
+go vet -all ./...
+rc=$((rc || $?))
 
-wantEmpty=$(golint ./...)
-if [ -n "$wantEmpty" ]; then
-  printf >&2 '\ngolint failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
-
-if [ -n "$failIt" ]; then
-  unset failIt
-  exit 1
-fi
-
+echo "Running go test"
 go test -v ./...
+rc=$((rc || $?))
 
+exit $rc


monopole:
Yeah, at first i copied the one-liner version used in many travis files too.

But i want a script that doesn't fail on the first error, and instead reports all problems with the PR that will make travis fail.  That's the point of the distinct tests.

Please just adjust for vendoring but leave overall behavior in place.

apelisse:
@monopole Updated PTAL. I've kept most of the former behavior, still simplified it IMHO.

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
 #!/bin/bash
 
-if [ -n "$failIt" ]; then
-  echo "Expecting failIt to be empty."
-  exit 1
-fi
+rc=0
```
hoegaarden: Is there a reason why we use `rc` to track return codes, rather than using `set -e` to bail out early on problems?

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
...
 
-wantEmpty=$(gofmt -s -d -l . 2>&1 )
-if [ -n "$wantEmpty" ]; then
-  printf >&2 'gofmt failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+go_dirs=$(go list -f {{.Dir}} ./...)
```
hoegaarden: - We're in the habit of using [shellcheck](https://www.shellcheck.net/) to lint our bash scripts. We'd like to quote `"{{.Dir}}"` rather than `{{.Dir}}` to [make our shellcheck editor plugin happy](https://github.com/koalaman/shellcheck/wiki/SC1083) :)
- We'd like to check the list of returned directories. If someone were to introduce a directory with whitespace in the directory name, then the tests below would fail with potentially mysterious error messages. We can check for this early, with something like: 
```sh
go_dirs=$(go list -f "{{.Dir}}" ./...)

for dir in $go_dirs
do
  [[ -d $dir ]] || {
    echo "Problem with directory $dir" >&2
    exit 1
  }
done
```

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
...
 
-wantEmpty=$(goimports -l .)
-if [ -n "$wantEmpty" ]; then
-  printf >&2 'goimports failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running go fmt"
+gofmt -s -d -l ${go_dirs}
+rc=$(($rc || $?))
```
hoegaarden: Shellcheck tells us that [it's generally better not to include the dollar in a regular variable name in an arithmetic expression](https://github.com/koalaman/shellcheck/wiki/SC2004). So we'd recommend changing `$(($rc || $?))` to `$((rc || $?))`

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
...
-  failIt=1
-fi
-
-if [ -n "$failIt" ]; then
-  unset failIt
-  exit 1
-fi
+echo "Running go vet"
+go vet -all ./...
+rc=$(($rc || $?))
```
hoegaarden: `$(($rc || $?))` --> `$((rc || $?))`

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
...
 
-wantEmpty=$(go vet -all ./... 2>&1)
-if [ -n "$wantEmpty" ]; then
-  printf >&2 'govet failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running goimports"
+diff -u <(echo -n) <(goimports -l ${go_dirs})
+rc=$(($rc || $?))
```
hoegaarden: `$(($rc || $?))` --> `$((rc || $?))`

bin/pre-commit.sh
```diff
@@ -1,37 +1,23 @@
...
-  unset failIt
-  exit 1
-fi
+echo "Running go vet"
+go vet -all ./...
+rc=$(($rc || $?))
 
+echo "Running go test"
 go test -v ./...
+rc=$(($rc || $?))
```
hoegaarden: `$(($rc || $?))` --> `$((rc || $?))`

apelisse:
> Since this is our first review, we're still learning how picky we should be -- please let us know if we're off the mark! We found a few minor things, which we mention in inline comments.

I'd rather have too picky than an ugly bug. I'll push back if I disagree anyway :-)

Thanks for the great review!

bin/pre-commit.sh
apelisse: Interesting! I don't think I new I could do that :-)

bin/pre-commit.sh
apelisse: Cool! Maybe we should run shellcheck as part of this script :-)

I'm glad you caught that. I was very aware of this though, just too lazy to fix it. Thank you for forcing me to do it :D 
I'll try to find a better fix though, we can probably do better with `xargs -0`.

bin/pre-commit.sh
apelisse: Yeah, as monopole mentionned earlier in this PR, we want to run all the tests, even if one fails. So that you don't have to fix-repush-look for failures repeatedly ...

bin/pre-commit.sh
hoegaarden: The question also is: Is it allowed to have whitespaces in directories/filenames, which means: is it valid to have whitespaces in package names?

apelisse: The problem is that `go list -f "{{.Dir}}"` returns the absolute path for each directory. Which means that if your code lives in "/home/my**SP**username/my-go-path/mypackagewithnospace", it's going to fail, but probably shouldn't

bin/pre-commit.sh
```diff
@@ -1,46 +1,25 @@
...
-go version
+rc=0
 
-# Assert state.
-if [ -n "$failIt" ]; then
-  echo "Expecting failIt to be empty."
-  exit 1
-fi
+go_dirs() {
+  go list -f '{{.Dir}}' ./... | tr '\n' '\0'
```
hoegaarden: I like that. The only thing which is not supported now are paths with `\n`, of course. But I think it is fair enough to not support that (at least for now).

bin/pre-commit.sh
```diff
@@ -1,46 +1,25 @@
...
+}
 
-wantEmpty=$(gofmt -s -d -l . 2>&1 )
-if [ -n "$wantEmpty" ]; then
-  printf >&2 '\ngofmt failed for:\n%s\n' "$wantEmpty"
-  failIt=1
-fi
+echo "Running go fmt"
+go_dirs | xargs -0 gofmt -s -d -l
```
hoegaarden: Observation: If any subdirectory has a `vendor` directory then this would also run `gofmt` on that - and will likely fail. I think that is fair, as we want to push for repo-global `vendor` anyway.

apelisse: Let's merge it that way, so that we can move forward, I'll try to come-up with something better :D

bin/pre-commit.sh
apelisse: Yeah, that'd be pretty unfortunate.

apelisse:
@monopole I'm merging this since this is blocking progress. I'll be happy to change it as you want when you're back.

PR #145, opened 2017-11-18
Tweak pre-commit to ignore vendor directory.

monopole:


apelisse:
I really don't see any reason why you don't get your pull-requests reviewed like everyone else. This might have solved two problems:
- Since you've re-copied parts of my PR into this one, I would have been happy to know that you did, instead of blindly try to address mine,
- I don't think this pull-requests fixes the initial problem anyway.

Thanks,
Antoine

Issue #87, opened 2017-10-19
Support integration test framework that brings up a new control plane


pwittrock:
Time commitment: 30-60 hours (assuming knowledge of Go and the Ginkgo test framework)

We need a way to write integration tests for kubectl and point it at a control plane (apiserver, controller-manager, etcd, scheduler) running locally.

As a kubectl developer, when I write an integration test for kubectl, there should be a ginkgo test framework that launches a new control plane, and provides a kubeconfig to the test both as a tmp file and as a datastructure.

Framework should take the control plane binaries as flags instead of compiling them from source.

See example of a test framework that brings up the control plane [here](https://github.com/kubernetes-incubator/apiserver-builder/blob/master/pkg/test/server.go).

[Link](https://k8s-testgrid.appspot.com/sig-cli-master#gke) to existing e2e tests

yland:
@pwittrock Could I be assigned to this issue, I would like to work on it.

pwittrock:
I won't be able to officially assign it to you until we add you as a collaborator, but will remove it from the open pool.  We can do this once you complete this first task.  @yland

pwittrock:
Please reach out if you later discover that you need more direction on this task.

pwittrock:
Also note that this task, while encapsulated, is a bit large in scope.  Feel free to tackle a smaller one for your first contribution instead.

pwittrock:
@yland unassigning since you picked up a different task

apelisse:
@pwittrock wrote somewhere else:

> **User story:**
> As a developer of CLIs that talk to Kubernetes APIs, I want to be able to write integration tests for my commands and libraries that run hermetically and locally in my development environment.  I want these tests to experience what happens when run against a real apiserver and control plane, but run fast and locally.
>
> **Solution:**
> Provide a test harness that will bring up the Kubernetes control plane locally (apiserver, etcd, controller-manager) locally for each tests.
>
> **Individual steps:**
> - Write a library to bring up etcd from a binary
> - Write a library to bring up apiserver from a binary, talk to the local etcd, and provide a kubeconfig file to talk to the apiserver from kubectl or client-go
> - Write a library to bring up the controller-manager from a binary and have it talk to the apiserver
> - Wrap the libraries in a test harness that will bring them up / tear them down for each test suite
> - Setup the pull / merge tests run against kubernetes/kubectl PRs so that the correct binaries are present and flags are provided to support the harness
[...]

PR #149, opened 2017-11-21
Beginnings of a test framework

totherme:
As discussed in https://github.com/kubernetes/kubectl/issues/87 and in slack and hangouts.

Known issues:
- If etcd starts slowly, the apiserver might not start in a sane state. This will be fixed soon.
- If anything is currently listening on any of the ports we use, there will be clashes.


apelisse:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubectl/149)

pkg/framework/test/integration/integration_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/kubectl/pkg/framework/test"
+)
+
+var _ = Describe("Integration", func() {
+	It("Successfully manages the fixtures lifecycle", func() {
+		assetsDir, ok := os.LookupEnv("KUBE_ASSETS_DIR")
```
apelisse: Are you planning on having a generic way to give the binaries? `KUBE_TEST_APISERVER` and `KUBE_TEST_ETCD`, so that my test can easily switch the underlying binary?

pkg/framework/test/integration/integration_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"path/filepath"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"k8s.io/kubectl/pkg/framework/test"
+)
+
+var _ = Describe("Integration", func() {
+	It("Successfully manages the fixtures lifecycle", func() {
```
apelisse: Would it make sense to have a benchmark test to see how long it takes to spawn and stop a fixture? I think that's a metric we'll be interested in.

pkg/framework/test/etcd.go
```diff
@@ -1,42 +1,54 @@
...
 
 // Start starts the etcd, and returns a gexec.Session. To stop it again, call Terminate and Wait on that session.
-func (e *Etcd) Start(etcdURL string, datadir string) error {
+func (e *Etcd) Start() error {
+	e.tempDirManager = &tempDirManager{}
+
+	dataDir := e.tempDirManager.Create()
+
 	args := []string{
+		"--debug",
```
apelisse: I have the feeling that this is going to be painful to maintain?

pkg/framework/test/temp_dir_manager.go
```diff
@@ -0,0 +1,49 @@
...
+type TempDirManager struct {
+	Maker   TempDirMaker
+	Remover TempDirRemover
+	dir     string
+}
+
+// Create knows how to create a temporary directory and how to keep track of it.
+func (t *TempDirManager) Create() (string, error) {
+	if t.dir == "" {
+		dir, err := t.Maker("", "kube-test-framework")
```
apelisse: `kube-test-framework-` ? (extra-dash)

pkg/framework/test/integration/integration_test.go
```diff
@@ -27,11 +27,11 @@ var _ = Describe("Integration", func() {
 
 		Eventually(func() bool {
 			return isSomethingListeningOnPort(2379)
-		}, 5*time.Second).Should(BeTrue(), "Expected Etcd to listen on 2379")
+		}, 25*time.Second).Should(BeTrue(), "Expected Etcd to listen on 2379")
```
apelisse: I foresee :snowflake:  :-)

.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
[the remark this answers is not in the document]
hoegaarden: Sure, we can remove that.

pkg/framework/test/apiserver.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// APIServer knows how to run a kubernetes apiserver. Set it up with the path to a precompiled binary.
+type APIServer struct {
+	// The path to the apiserver binary
+	Path string
+}
+
+// Start starts the apiserver, and returns a gexec.Session. To stop it again, call Terminate and Wait on that session.
+func (s APIServer) Start(etcdURL string) (*gexec.Session, error) {
```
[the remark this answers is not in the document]
hoegaarden: - Right now only the etcd port is configurable. However we know that we might need to make other ports configurable or random. We have [some](https://www.pivotaltracker.com/story/show/153178204) [stories](https://www.pivotaltracker.com/story/show/153178221) for that in our backlog to address that and see if and how we can make all the components run multiple times in parallel and play nicely together.
- We started tracking the time it takes to bring up and bring down the whole thing, see [here](https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/23#L5a1d829b:71).
- I see your point with the incompatible options. When this becomes an issue we can introduce some versioning or something else to address that.

pkg/framework/test/apiserver_test.go
```diff
@@ -0,0 +1,51 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
+	"github.com/onsi/gomega/gexec"
+)
+
+var _ = Describe("Apiserver", func() {
+
+	Context("when given a path to a binary that runs for a long time", func() {
+		It("can start and stop that binary", func() {
+			pathToFakeAPIServer, err := gexec.Build("k8s.io/kubectl/pkg/framework/test/assets/fakeapiserver")
```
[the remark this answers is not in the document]
hoegaarden: When you say "the coupling with `main.go` below is kind of awkward", we think you're primarily referring to the duplication between the arguments in [test/apiserver.go](https://github.com/totherme/kubectl/blob/6c4975851eabe176923b978af0ca77097305b3f5/pkg/framework/test/apiserver.go#L41-L51) and [fakeapiserver/apiserver.go](https://github.com/totherme/kubectl/blob/6c4975851eabe176923b978af0ca77097305b3f5/pkg/framework/test/assets/fakeapiserver/apiserver.go#L12-L23). We totally agree that this is awkward.

That fake is only used by this one test -- so if we delete this test, we can delete that fake. That might be a good thing -- especially since the wiring which that test was originally written to drive out is now covered by a higher level integration test. The non-happy-path cases in this file cover code that's not covered anywhere else, so we'd definitely like to keep those. 

From a "tests as docs" viewpoint, it feels like a little bit of a pity to leave that file as an incomplete doc -- describing only the error cases... But maybe that's actually just fine? We might have a bit of a think about whether we can write any useful happy-path test (motivated primarily by wanting to keep some executable docs for that path) that doesn't use that fake.

pkg/framework/test/apiserver_test.go
```diff
@@ -0,0 +1,51 @@
...
+			apiServer := APIServer{}
+			_, err := apiServer.Start("the etcd url")
+			Expect(err).To(MatchError(ContainSubstring("no such file or directory")))
+		})
+	})
+
+	Context("when given a path to a non-executable", func() {
+		It("fails with a helpful error", func() {
+			apiServer := APIServer{
+				Path: "./apiserver.go",
```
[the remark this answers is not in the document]
hoegaarden: Good point -- we can use a temporary file here.

Is this also an issue with any use of `gexec.Build` ?

pkg/framework/test/etcd.go
hoegaarden: If this is only about using `--debug` or not, we can totally remove that line. If it's about management of arguments (potentially for different versions, ...), we believe we have already covered that in [the other comment](https://github.com/kubernetes/kubectl/pull/149#discussion_r153734672).

pkg/framework/test/fixtures.go
```diff
@@ -0,0 +1,83 @@
+package test
```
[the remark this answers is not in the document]
hoegaarden: A couple of random thoughts we just discussed:

- We tried to make the interface for the processes we start in our framework rather generic. That should make it relatively easy to introduce new processes (Scheduler, Controller(Manager), ...) somewhen in the future.
- We see that the APIServer depends on Etcd -- from that point of view it might be a good idea to make this dependency explicit as you suggested.
- I believe we can run the APIServer without Etcd. When Etcd eventually comes up APIServer will connect and everything will just work. We might have test which actually test with a missing Etcd and/or an dis- and reappearing Etcd.
- ( What if I'd like to run my tests against a already running instance of Etcd on a different machine or so?[1] )

So we are not against making the dependency explicit, we think both ways have their pros and cons. I -- personally -- would opt for keeping it as is for now, but that's just a gut feeling. We are happy to discuss that in more depth!


[1] That just came to our minds -- we know that is not supported or even planned to be supported by our framework. However we thought that might be a valuable use case to keep in mind ...

pkg/framework/test/integration/integration_test.go
hoegaarden: Yeah ... we need to figure out how to handle that properly -- we dropped [a story](https://www.pivotaltracker.com/story/show/153245508) in our backlog for that.

pkg/framework/test/integration/integration_test.go
hoegaarden: We introduced that in ["Track the time it takes to start&stop the fixtures"](https://github.com/totherme/kubectl/commit/a0e9df8b57e5c0bff3a4e4ef3f29f82dd0290cdc)

You can find the timings in our pipeline:
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/23#L5a1d829b:71
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/22#L5a1d90c7:27
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/20#L5a1d8f04:71
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/16#L5a1558c6:24
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/15#L5a1558c3:20
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/14#L5a15589a:19
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/13#L5a1558b8:62 (first test with parallel start)
- https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl/jobs/test-dev-branch/builds/12#L5a1558b5:65 (last test with serial start)

pkg/framework/test/integration/integration_test.go
hoegaarden: Yes. We have that in the back of our heads. We have [one story](https://www.pivotaltracker.com/story/show/153178125) in our backlog for that.

pkg/framework/test/ci/pipeline.yml
```diff
@@ -0,0 +1,70 @@
+
```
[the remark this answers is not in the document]
hoegaarden: This is just the [concourse pipeline](https://wings.concourse.ci/teams/k8s-c10s/pipelines/kubectl) config to test the framework and merge our changes into the PR branch if the tests go green.

We will see if we can [make the same/similar thing happen](https://www.pivotaltracker.com/story/show/153245975) with Travis or if we can have the concourse stuff somewhere else. We just went with what we know and used to here but happy to change to something else.

pkg/framework/test/integration/integration_test.go
hoegaarden: BTW: You can run our tests via `./scripts/run-tests.sh`; If you run `GINKGO_PERFORMANCE=1 ./scripts/run-tests.sh` it will also do some benchmarking.

pkg/framework/test/democli/cmd/listPods.go
```diff
@@ -0,0 +1,65 @@
...
+	Use:   "listPods",
+	Short: "List all pods",
+	Long:  `Give a list of all pods known by the system`,
+	Run: func(cmd *cobra.Command, args []string) {
+		runGetPods()
+	},
+}
+
+func runGetPods() {
+	config, _ := clientcmd.BuildConfigFromFlags("http://localhost:8080", "")
```
apelisse: Should this be a parameter instead?

pkg/framework/test/democli/integration/integration_test.go
```diff
@@ -20,4 +20,11 @@ var _ = Describe("Integration", func() {
 		Expect(session.Out).To(gbytes.Say(helpfulMessage))
 	})
 
+	It("can get a list of pods", func() {
+		command := exec.Command(pathToDemoCommand, "listPods")
```
apelisse: APIServer coordinates should be passed here.

pkg/framework/test/integration/integration_test.go
```diff
@@ -40,6 +32,15 @@ var _ = Describe("Integration", func() {
 		By("Ensuring APIServer is not listening anymore")
 		Expect(isAPIServerListening()).To(BeFalse(), "Expected APIServer not to listen anymore")
 	})
+
+	Measure("It should be fast to bring up and tear down the fixtures", func(b Benchmarker) {
```
apelisse: Cool! I didn't know that!

pkg/framework/test/etcd.go
```diff
@@ -18,11 +18,10 @@ type Etcd struct {
 	session        *gexec.Session
 	stdOut         *gbytes.Buffer
 	stdErr         *gbytes.Buffer
-	dataDirManager DataDirManager
+	dataDirManager dataDirManager
 }
 
-// DataDirManager knows how to create and destroy Etcd's data directory.
-type DataDirManager interface {
+type dataDirManager interface {
```
apelisse: Or you can move that to an internal packages if you want to reuse it somewhere else.

pkg/framework/test/assets/fakeapiserver/apiserver.go
```diff
@@ -3,21 +3,24 @@ package main
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"time"
 )
 
 func main() {
-	expectedArgs := []string{
```
apelisse: This list is going to be a nightmare to maintain. Isn't it already?

bin/pre-commit.sh
```diff
@@ -19,7 +19,12 @@ go vet -all ./...
 rc=$((rc || $?))
 
 echo "Running go test"
-go test -v ./...
+go list ./... | grep -vF pkg/framework/test | xargs go test -v
```
apelisse: I think there should be some sort of "default" value for `KUBE_ASSERT_DIR` and maybe we can download the binaries before running `go test`?

pkg/framework/test/integration/integration_test.go
apelisse: That's super exciting to see! The numbers are not that exciting. Let's see later what we can do.

bin/pre-commit.sh
hoegaarden: Agreed. I think this also ties back to making our think work well with `go test`. As soon as we have that we can retire our test wrapper script `pkg/framework/test/scripts/run-tests.sh` and think about better ways of handling the test binaries/assets resp. their location. We in fact have all those things in our backlog.

pkg/framework/test/democli/cmd/listPods.go
hoegaarden: Definitely. We wont to have our framework generate some sort of config (the framework will be the thing that determines the port eg. the APIServer is listening on) and somehow expose that to the clients resp. the component under test. Until we have that and somewhat in the spirit of TDD we just did the simplest thing possible to make our stuff work. Next iterations will most definitely change that.

totherme:
Thanks @apelisse for all your comments! We've opened issues to track your two main concerns: #152 and #153 . We'll start work on those now.

Issue #153, opened 2017-11-30
The Test Framework's test fakes shouldn't be so tightly coupled with the implementations they test


totherme:
In the test framework MVP (discussed in #87 ), there is currently a tight coupling between the test fakes `pkg/frameworks/test/assets/fakeetcd/` and `etcd.go`. In particular, both of these currently know the exact flags that `etcd` ought to be called with. There is a similar issue with the apiserver fake.

We should find a way to adequately test this wiring which doesn't leave us with tightly coupled test-doubles like this. This might involve leaning more heavily on the integration tests for the happy path, and removing the unit-level happy path tests.

This issue began as a conversation with @apelisse in #149

totherme:
After investigating, we think that fixing this issue will make #152 easier, so we'd like to start with this.

@hoegaarden and I would like to work on this please.

cc @seans3 @mengqiy I would like to take this

hoegaarden:
As #154 & #155 are merged, this issue can be closed.
@totherme @apelisse

apelisse:
I think you can use `/close` or something :D

PR #154, opened 2017-12-04
Proper fake for etcd

hoegaarden:
@apelisse , this is the first part of us removing this fake binaries and thus the tight coupling as described in https://github.com/kubernetes/kubectl/issues/153 .


pkg/framework/test/etcd.go
```diff
@@ -12,10 +13,10 @@ import (
 
 // Etcd knows how to run an etcd server. Set it up with the path to a precompiled binary.
 type Etcd struct {
-	// The path to the etcd binary
 	Path           string
 	EtcdURL        string
-	session        *gexec.Session
+	ProcessStarter simpleSessionStarter
```
apelisse: These changes seem to always be the same. How much more complex would it be to share some of this code between `apiserver` and `etcd`. And I can forsee other components in the future, will you have to duplicate this code again and again?

hoegaarden: That is indeed true. I can check tomorrow if we can unify them.

PR #155, opened 2017-12-04
Proper fake for apiserver

hoegaarden:
@apelisse , this is the second part of us removing this fake binaries and thus the tight coupling as described in #153 .

Related to #154 .


pkg/framework/test/etcd_test.go
```diff
@@ -51,18 +55,36 @@ var _ = Describe("Etcd", func() {
...
 			etcd.Stop()
 			Expect(etcd).To(gexec.Exit(143))
 			Expect(fakeSession.TerminateCallCount()).To(Equal(1))
 			Expect(fakeSession.WaitCallCount()).To(Equal(1))
 			Expect(fakeSession.ExitCodeCallCount()).To(Equal(2))
+			Expect(fakeDataDirManager.DestroyCallCount()).To(Equal(1))
+		})
+	})
+
+	Context("when the data directory cannot be crreated", func() {
```
apelisse: typo :D

pkg/framework/test/apiserver_test.go
```diff
@@ -1,62 +1,106 @@
...
+			Expect(fakeSession.ExitCodeCallCount()).To(Equal(2))
+			Expect(fakeCertDirManager.DestroyCallCount()).To(Equal(1))
 		})
-
 	})
 
-	Context("when no path is given", func() {
-		It("fails with a helpful error", func() {
-			apiServer := &APIServer{}
+	Context("when the certificate directory cannot be crreated", func() {
```
apelisse: typo again :-)

pkg/framework/test/etcd.go
```diff
@@ -12,10 +13,10 @@ import (
 
 // Etcd knows how to run an etcd server. Set it up with the path to a precompiled binary.
 type Etcd struct {
-	// The path to the etcd binary
 	Path           string
 	EtcdURL        string
-	session        *gexec.Session
+	ProcessStarter simpleSessionStarter
```
apelisse: I find it odd to have the "ProcessStarter" as a member of the object, rather than the other way around (I would expect the ProcessStarter to take the object to start as an argument).

hoegaarden: I see your point. To be honest, we never thought about it that way. Let me think about that, and discuss that with @totherme tomorrow.

PR #157, opened 2017-12-05
No special treatment for test framework

### LINKED ISSUE #152: The Test Framework should not require special treatment in pre-commit.sh
There is an MVP of the test framework discussed in #87 , but it relies on having special treatment in pre-commit.sh to pass the travis tests. In particular:

- The test framework directory is currently exempt from having to run `go test ./...`
- Instead, a custom script `pkg/framework/test/scripts/run-tests.sh` is called, which runs the tests using the ginkgo CLI

The tests of the test framework should be changed so as not to require this special treatment. At first glance, this will involve:

- Changing the framework to bring up etcd and the apiserver on random unused ports, instead of the default ones
- Having sensible defaults for all environment variables that the framework relies on.

This issue began as a conversation with @apelisse in #149

hoegaarden:
Based on top of
- #154 
- #155 

As discussed with @apelisse .

[closes #152]


pkg/framework/test/integration/integration_suite_test.go
```diff
@@ -23,8 +24,18 @@ var (
 )
 
 var _ = BeforeSuite(func() {
-	assetsDir, ok := os.LookupEnv("KUBE_ASSETS_DIR")
-	Expect(ok).To(BeTrue(), "Expected $KUBE_ASSETS_DIR to be set")
+	assetsDir := ""
+
+	if dirFromEnv, ok := os.LookupEnv("KUBE_ASSETS_DIR"); ok {
```
apelisse: I'm still not a big fan of `KUBE_ASSETS_DIR` rather than individual environment variable for each binary.

bin/pre-commit.sh
```diff
@@ -18,14 +26,12 @@ echo "Running go vet"
...
+echo "Installing test binaries"
+./pkg/framework/test/scripts/download-binaries.sh
 rc=$((rc || $?))
 
-echo "Running test framework tests"
-go get github.com/onsi/ginkgo/ginkgo \
-  && ./pkg/framework/test/scripts/download-binaries.sh \
-  && ./pkg/framework/test/scripts/run-tests.sh
+echo "Running go test"
+go test -v ./...
```
apelisse: yay :-)

pkg/framework/test/apiserver.go
```diff
@@ -33,24 +31,22 @@ type certDirManager interface {
...
 		CertDirManager: NewTempDirManager(),
-		APIServerURL:   apiServerURL,
 	}
 
 	return apiserver
 }
 
 // Start starts the apiserver, waits for it to come up, and returns an error, if occoured.
-func (s *APIServer) Start() error {
+func (s *APIServer) Start(config map[string]string) error {
```
apelisse: I'm not a very big fan of the map[string]string for configuration. I understand that we can't pass in a specific structure because `Start` is part of an interface, but then can we pass in the options at creation time? That's a big tell that something is going wrong here I think.

pkg/framework/test/etcd.go
```diff
@@ -62,7 +58,7 @@ func NewEtcd(pathToEtcd, etcdURL, etcdPeerURL string) *Etcd {
 }
 
 // Start starts the etcd, waits for it to come up, and returns an error, if occoured.
-func (e *Etcd) Start() error {
+func (e *Etcd) Start(config map[string]string) error {
```
apelisse: Ditto

pkg/framework/test/etcd.go
```diff
@@ -45,15 +43,13 @@ type SimpleSession interface {
 type simpleSessionStarter func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error)
 
 // NewEtcd constructs an Etcd Fixture Process
-func NewEtcd(pathToEtcd, etcdURL, etcdPeerURL string) *Etcd {
+func NewEtcd(pathToEtcd string) *Etcd {
```
apelisse: I was going to note on the other commit before that I was a little bit concerned about the handling of configurations here, I'm not sure I agree with the direction.

pkg/framework/test/fixtures.go
```diff
@@ -24,62 +24,71 @@ type FixturesConfig struct {
...
+	etcdPeerURL, err := getHTTPListenURL()
+	if err != nil {
+		return err
+	}
+	apiServerURL, err := getHTTPListenURL()
+	if err != nil {
+		return err
+	}
+
+	etcdConf := configs{
```
apelisse: Can this be moved to the fixture creation?

pkg/framework/test/apiserver.go
hoegaarden: I understand that `map[string]string` is not really good. I went with that, because it was an easy first step. Anyway, I was thinking about some sort of `FixtureConfig` interface, that one can pass to `Start` and which is implemented by an `EtcdConfig` and an `APIServerConfig`.

pkg/framework/test/etcd.go
hoegaarden: I am not sure what you want to tell me with the comment. Is it any of those:
- All configuration should be given to the constructor.
- Not even the path to the binary should be given at construction time but configured later.

My line of thought was something like:
- If I have, for example, 2 different etcd binaries, my `Etcd`s will be different things. Therefore `Etcd` will be constructed with a specific binary.
- The configuration might change, but as long as the binary stays the same, it is the very same `Etcd`. The configuration might also change between different runs. Therefore to place to hand that in is actually at start time.

pkg/framework/test/etcd.go
hoegaarden: DItto ;)

hoegaarden:
Thanks for your comments!


There are a couple of things I'd like to discuss because I am not sure what you mean and/or I am not sure if I understand things correctly:

> I have the feeling that if Fixtures wasn't keeping everything together, it would fall apart, which worries me a little bit.

I am not sure what you picture the purpose f the `Fixtrues`. From my point of view, that's *the thing* that brings up the whole testing framework, you give it a path to an Etcd & and APIServer and it knows how to run those. For users of the `Fixtures` everything is hidden for them, except (for now, in future more such things ...) the APIServer's URL they can test their client against.
So, the whole purpose of the `Fixtures` is keeping everything together.

> I was going to note on the other commit before that I was a little bit concerned about the handling of configurations here, I'm not sure I agree with the direction.

Some reasoning why I handed in the config on the `Start` and not on construction time can be found [in the other comment](https://github.com/kubernetes/kubectl/pull/157#discussion_r155184058). Having said that, I am fine with pushing the configuration handling back to the constructors. To make things nicer, I introduced config objects.
To be honest, I am not entirely sure if that is more the direction you want to go ... ?

Would be great if we could discuss those, and any more comments you might have on that PR!

Thanks!

apelisse:
OK I think this is better, I like the configuration objects much better.
I'll get that merged, but I'll leave that here for future references:
- I think the `Fixture` is doing to much, and I'd rather live in a world where it doesn't exist at all, because it's not needed.
- In order to do that, I think we would have to improve the interface of the APIServer and Etcd server a little bit. Interfaces are not bad per-se but:
   - Existing configuration requires all values to be specified, I think it'd be nice to have sensible defaults,
   - Configuring the binaries path is kind of awkward and is not super easy to manage. I'm not sure how we could improve that, but I'm sure we can make that more composable.
- We will certainly want something like `DefaultControlPlane` (or `Fixture` to keep the current naming), but I think this should only be an after-thought.
```
// I want default behaviors
APIServer{}.Start()
Etcd{}.Start()
// Yay!

// I want non-default behavior:
APIServer{
   SomeFake: MyFake,
   SomeConfig: Config,
   APIServerBinary: BinaryFinder("KUBETEST_APISERVER_BINARY"),
}.Start()
Etcd{
   // Other config
}.Start()
```

hoegaarden:
Thanks for the comments @apelisse !

We've opened issues for sane 0-value behaviour: #161, #162, and #163. Getting sane 0-value behaviour for `Etcd` and `Fixtures` (#161 and #163) seems straightforward. Getting sane 0-value behaviour for `APIServer` (#162) seems to us to drive to the heart of your previous suggestions about making the dependency on Etcd more explicit. It also seems to us to be potentially a little trickier, with trade-offs for things like start-times and consistency of approach among our various fixtures. We'll get started on #161 and #163, and would love to chat about #162 in more detail if you're interested.

apelisse:
Thank you guys! Happy to see the `go test ./...` back to normal :-)

PR #158, opened 2017-12-05
Move `openapi` package from `k8s.io/kubernetes` to `k8s.io/kubectl`

### LINKED ISSUE #137: Move `k8s.io/kubernetes/pkg/kubectl/cmd/util/openapi` to `k8s.io/kubectl/pkg/openapi`
Or maybe it should go somewhere else, to be thought through.

AlexKhotian:
- Moved packages openapi and validation;
- Improved  a bit the documentation of openapi package;
- Added examples for validation and openapi_getter;
- Cherry-picked a swagger.json for tests(validation_test and openapi_test)

Fixes #137


AlexKhotian:
Add a CLA sign.
/assign @apelisse

pkg/openapi/validation/example_validation_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+package validation
+
+import (
+	"fmt"
+
+	"k8s.io/kubectl/pkg/openapi"
+)
+
+func GetResources() (openapi.Resources, error) {
```
apelisse: Can we get rid of this somehow? It's only used once. And don't we have a fake that creates the `resources` directly?

pkg/openapi/example_openapi_getter_test.go
```diff
@@ -0,0 +1,38 @@
...
+limitations under the License.
+*/
+
+package openapi_test
+
+import (
+	"fmt"
+
+	"k8s.io/kubectl/pkg/openapi"
+	tst "k8s.io/kubectl/pkg/openapi/testing"
```
apelisse: When I wrote this code code, I didn't know that the convention seems to be "openapitest" rather than anything else. e.g. [httptest](https://golang.org/pkg/net/http/httptest/). Would you mind renaming that package?

pkg/openapi/openapi_getter.go
```diff
@@ -0,0 +1,65 @@
...
+	openAPISchema Resources
+	err           error
+
+	openAPIClient discovery.OpenAPISchemaInterface
+}
+
+var _ Getter = &synchronizedOpenAPIGetter{}
+
+// Getter is an interface for fetching openapi specs and parsing them into an Resources struct
+type Getter interface {
```
apelisse: I'm starting to get a little bit confused about all these interfaces, and it looks like the code is using a different method every single time. We have the client, the getter, the resources, the OpenAPISchemaInterface. Do we need all of these? It also complicates the "openapi/testing" code that needs to implement many variations.

apelisse:
For reference, this is addressing #137

mengqiy:
Overall LG.
But I'd like to see the git history of the moving package is retained. And your change is applied on top of it.

apelisse:
> But I'd like to see the git history of the moving package is retained. And your change is applied on top of it.

We should discuss about that. I don't really see the value to it compared for the cost. We could get the same by adding a link to the origin location.

mengqiy:
Let's try to retain the git history, if you encounter quite some hard problem please let us know.

apelisse:
> Let's try to retain the git history, if you encounter quite some hard problem please let us know.

Perfect conclusion!

apelisse:
@AlexKhotian I'm renaming this PR and adding the issue number to the body message. It's not very obvious right now :-). Feel free to rename as you wish later.

pkg/openapi/openapi_getter.go
AlexKhotian: Can it be done as separate step? We can discuss more detailed, what and why is needed. Perhaps have some additional issue.

apelisse: Yes, absolutely!

Issue #136, opened 2017-11-15
Move `k8s.io/kubernetes/pkg/kubectl/explain` to `k8s.io/kubectl/pkg/explain`


apelisse:
And make sure it's properly tested, documented, with nice examples, few dependencies.

wackxu:
I can help do this.

apelisse:
@wackxu great! Let me know if you need any feedback/help or anything else. Feel free to ping me on slack (apelisse) if you want IM. Thank you!

apelisse:
I can't assign to you since you're not "part of the org". I'm assigning to me to make it clear that someone is working on it.

mengqiy:
> I can't assign to you since you're not "part of the org".

Do we have policy about this?

I guess you can assign using github instead of using `/assign`

apelisse:
No :-/ I just can't find that name in the list.

mengqiy:
I see. Interesting :-\

PR #160, opened 2017-12-06
[WIP] Move `k8s.io/kubernetes/pkg/kubectl/explain` to `k8s.io/kubectl/pkg/explain

wackxu:
fix https://github.com/kubernetes/kubectl/issues/136


apelisse:
Some comments on that PR:
- Thank you for keeping that history, I think it's useful. Maybe we want to connect that to the rest of the tree (by grafting the parent commit of the first commit). I can potentially help with that,
- If we connect the trees, we don't have to get the merge commit in that PR, I'd like to avoid that,
- I don't think we want to keep the `BUILD` files,
- We absolutely don't want to import packages from `k8s.io/kubernetes`,
- Yes, you need to import `k8s.io/kubernetes` to move that package, most likely because this actually depends on #158 first (sorry I didn't notice that before),
- We don't want to keep the same path, i.e. `pkg/kubectl/...` but `pkg/...`, since kubectl is in the name of the repository,
- @mengqi, can we avoid importing all the files into `/` and then move them to their actual location? What `filter-branch` command did you use before?

wackxu:
@apelisse Fix 3, 6, `cmdtesting "k8s.io/kubernetes/pkg/kubectl/cmd/testing"` also make us use import from `k8s.io/kubernetes`, we may need move more packages to not use `k8s.io/kubernetes` or just use it until we move other packages here.

wackxu:
@apelisse I am sorry I do not know how to deal with 1, 2

wackxu:
I use command  `git filter-branch -f --prune-empty --subdirectory-filter pkg/kubectl/explain/` when retain the history

pkg/explain/OWNERS
```diff
@@ -0,0 +1,4 @@
+approvers:
```
apelisse: We don't use that yet. Feel free to remove that file.

pkg/explain/typename_test.go
```diff
@@ -0,0 +1,80 @@
...
+limitations under the License.
+*/
+
+package explain
+
+import (
+	"testing"
+
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	tst "k8s.io/kubernetes/pkg/kubectl/cmd/util/openapi/testing"
```
apelisse: No `k8s.io/kubernetes`! That should be fixed by #158 though

pkg/explain/explain_test.go
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+package explain
+
+import (
+	"reflect"
+	"testing"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	cmdtesting "k8s.io/kubernetes/pkg/kubectl/cmd/testing"
```
apelisse: We can't import that. Commenting below ...

pkg/explain/explain_test.go
```diff
@@ -0,0 +1,81 @@
...
+		if !reflect.DeepEqual(test.expectedFieldsPath, gotFieldsPath) && !test.expectedErr {
+			t.Errorf("%s: expected fieldsPath: %s, got: %s", test.name, test.expectedFieldsPath, gotFieldsPath)
+		}
+	}
+}
+
+func getMapper() meta.RESTMapper {
+	f, _, _, _ := cmdtesting.NewTestFactory()
+	mapper, _ := f.Object()
+	return mapper
```
apelisse: All we need is the mapper. Do we have to create a factory to retrieve the object? Could you please look into it?

mengqiy:
> can we avoid importing all the files into / and then move them to their actual location?

Maybe we can try to do something similar to https://github.com/kubernetes/test-infra/pull/4134/files#diff-01152b7c951c51b648bcb760d34df873?

> What filter-branch command did you use before?

I followed this last time. http://gbayer.com/development/moving-files-from-one-git-repository-to-another-preserving-history/

pkg/explain/explain_test.go
mengqiy: One option is to use construct the mapper from `Registry` https://github.com/kubernetes/kubernetes/blob/a505d1bc7854c0581fc8dcbc007ee0063d0fdfb3/pkg/kubectl/scheme/scheme.go#L34
Then we need to move Scheme and Registry to kubectl repo.

wackxu:
@apelisse As @mengqiy say, 

1. move `scheme` and `registry` to remove dependency from `k8s.io/kubernetes`,  

2. update import when #158 get merged then can merge this.

3. update vendor because `registry` need new denpendency for group `admission` https://github.com/kubernetes/kubernetes/pull/56004  
```The `admission/v1alpha1` API has graduated to `v1beta1`. Please delete your existing admission webhooks before upgrading the cluster, and update them to use the latest API, because the API has backwards incompatible changes.```

apelisse:
It makes me really really sad that we need to import `scheme` and `registry`, update tons of versions of API, just because we want a function that `singularizes` in a test ... Can we refactor a little bit so that we don't have to depend on both of these?

PR #165, opened 2017-12-08
add gitignore file

wackxu:
/assign @apelisse


apelisse:
I don't know about any of these files. If your OS/tools are leaving traces everywhere, please refer to [gitignore documentation](https://git-scm.com/docs/gitignore#_description), I think this belongs to `$GIT_DIR/info/exclude` or `$HOME/.config/git/ignore` (depending on `core.excludesFile` for the later)

mengqiy:
@apelisse I have a different opinion of this PR.
I think we should do something similar as https://github.com/kubernetes/kubernetes/blob/master/.gitignore. IMHO project-based gitignore is more convenient than the user-based one.

apelisse:
yeah, if that's what Kubernetes does :-)
Everyone has it's own IDE, and I don't think we should update the gitignore anytime somebody starts using a new one. But sure.

mengqiy:
@wackxu IMO, gitignore can have at least the following. Can you update the PR?

```
# OSX leaves these everywhere on SMB shares
._*

# OSX trash
.DS_Store

# Eclipse files
.classpath
.project
.settings/**

# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
.idea/
*.iml

# Vscode files
.vscode

# This is where the result of the go build goes
/output*/
/_output*/
/_output

# Emacs save files
*~
\#*\#
.\#*

# Vim-related files
[._]*.s[a-w][a-z]
[._]s[a-w][a-z]
*.un~
Session.vim
.netrwhist

# cscope-related files
cscope.*

# Mercurial files
**/.hg
**/.hg*
```

wackxu:
@mengqiy Updated

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+*.un~
+Session.vim
+.netrwhist
+
+# cscope-related files
+cscope.*
+
+# Mercurial files
+**/.hg
+**/.hg*
```
mengqiy: We'd better have a newline here.

.gitignore
```diff
@@ -0,0 +1,41 @@
...
+.settings/**
+
+# Files generated by JetBrains IDEs, e.g. IntelliJ IDEA
+.idea/
+*.iml
+
+# Vscode files
+.vscode
+
+# This is where the result of the go build goes
```
mengqiy: These 4 lines don't apply to this repo ATM. Please remove.

Issue #162, opened 2017-12-07
In the testing framework, the APIServer struct should have sane 0-value behaviour


hoegaarden:
In #157 @apelisse commented:

```
// I want default behaviors
APIServer{}.Start()
Etcd{}.Start()
// Yay!
```

That doesn't currently work. Let's make the APIServer bit of it it work. (the etcd bit is in #161, and 
making it work for `Fixtures{}` is #163)

More specifically:

`&APIServer{}.Start()` should start an etcd server listening on some random port, then start an api server that talks to that etcd. The api server should be listening on a different random port. We ought to be able to query the APIServer struct for the api server port.

### Runtime dependancy on etcd

We think that for a 0-value APIServer to start successfully, one of the following three cases needs to hold:
  1. The api server binary can run without talking to etcd. This is not the case.
  2. The APIServer struct default behaviour is to connect its api server binary to an etcd running on the default port. Testing this with `go test ./...` is tricky, because those tests all run concurrently, and ports can clash. See the work in #157 to remove our dependency on a default etcd port.
  3. The APIServer struct could be responsible for starting etcd as well as the api server. This way, we can get the port that etcd actually starts on, and connect to that.

Of these three options, only option (3) seems tenable. If we go down that route, this means that  something like the Fixtures struct can no-longer be responsible for starting the api server and etcd in parallel. Whatever the APIServer does to start etcd is how it has to be.

### Possible future dependancies

If we find that we want to add other services that the APIServer needs to talk to (such as a controller manager, or a fake scheduler, or whatever), then we might hit something like this issue again:
  1. The api server binary can run without talking to the controller manager. This *is* the case.
  2. The APIServer struct default behaviour is to connect its api server binary to a controller manager running on the default port. Testing this with `go test ./...` is tricky, because those tests all run concurrently, and ports can clash.
  3. The APIServer struct could be responsible for starting the controller manager as well as the api server. This way, we can get the port that the controller manager actually starts on, and connect to that.

If we choose option (3) this time, then the APIServer becomes something like the current implementation of the Fixtures, coordinating processes, but in addition to that also starting the api server process. It seems that for all non-etcd future dependencies, we should choose option 1.

PR #166, opened 2017-12-08
Add a sane default constructor for the Etcd and EtcdConfig structs

### LINKED ISSUE #161: In the testing framework, the Etcd struct should have sane 0-value behaviour
In #157 @apelisse commented:

```
// I want default behaviors
APIServer{}.Start()
Etcd{}.Start()
// Yay!
```

That doesn't currently work. Let's make the Etcd bit of it it work. (the apiserver bit is in #162, and 
making it work for `Fixtures{}` is #163)

More specifically:

`&Etcd{}.Start()` should start an etcd server listening on some random port. We ought to be able to query the struct for that port.

totherme:
[closes #161]

We need to do some calculation to find binary paths and free ports, so we can't make a 0-value Etcd struct useful. But we can provide a constructor that takes no arguments, and gives you good defaults.

Cc @hoegaarden @apelisse

### DIFF
--- a/pkg/framework/test/etcd.go
+++ b/pkg/framework/test/etcd.go
@@ -43,8 +43,29 @@ type SimpleSession interface {
 
 type simpleSessionStarter func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error)
 
-// NewEtcd constructs an Etcd Fixture Process
-func NewEtcd(pathToEtcd string, config *EtcdConfig) *Etcd {
+var etcdBinPathFinder = DefaultBinPathFinder
+
+// NewEtcd returns a Etcd process configured with sane defaults
+func NewEtcd() (*Etcd, error) {
+	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
+		return gexec.Start(command, out, err)
+	}
+
+	config, err := NewEtcdConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return &Etcd{
+		Path:           etcdBinPathFinder("etcd"),
+		ProcessStarter: starter,
+		DataDirManager: NewTempDirManager(),
+		Config:         config,
+	}, nil
+}
+
+// NewEtcdWithBinaryAndConfig returns a Etcd process, using the handed in binary and config.
+func NewEtcdWithBinaryAndConfig(pathToEtcd string, config *EtcdConfig) *Etcd {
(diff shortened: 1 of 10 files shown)


totherme:
/assign @apelisse 

:D

pkg/framework/test/etcd.go
```diff
@@ -43,8 +43,29 @@ type SimpleSession interface {
...
+	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
+		return gexec.Start(command, out, err)
+	}
+
+	config, err := NewEtcdConfig()
+	if err != nil {
+		return nil, err
+	}
+
+	return &Etcd{
```
apelisse: Why don't we delay defaulting until `Start` is called?

hoegaarden: Just because we wanted to refactor "all the things" (esp. #162) first. We can change that once this is done.

pkg/framework/test/etcd_config_test.go
```diff
@@ -1,36 +1,68 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(conf.ClientURL).To(Equal("http://1.2.3.4:42"))
+		})
 
-	It("errors on empty config", func() {
-		conf := &EtcdConfig{}
-		err := conf.Validate()
-		Expect(err).To(MatchError(ContainSubstring("PeerURL: non zero value required")))
-		Expect(err).To(MatchError(ContainSubstring("ClientURL: non zero value required")))
+		It("prop[agates and error while trying to find a port", func() {
```
apelisse: typo :-)

pkg/framework/test/etcd_config_test.go
```diff
@@ -1,36 +1,68 @@
...
+		)
+		BeforeEach(func() {
+			previousPortFinder = etcdPortFinder
+		})
+		AfterEach(func() {
+			etcdPortFinder = previousPortFinder
+		})
+
+		It("sets some sane default URLs", func() {
+			etcdPortFinder = func(host string) (port int, addr string, err error) {
```
apelisse: Yeah, that's a little unfortunate. I see that as a sign that we're doing something wrong. Do you know what it is?

apelisse: We might want to pass in the portfinder, with it being optional? Might be related to my earlier comment.

PR #168, opened 2017-12-11
Finalize the interface of the Fixtures struct

### LINKED ISSUE #163: In the testing framework, the Fixtures struct should have sane 0-value behaviour
In #157 @apelisse commented:

```
// I want default behaviors
APIServer{}.Start()
Etcd{}.Start()
// Yay!
```

That doesn't currently work for either of those structs, or even for the `Fixtures{}` struct. Let's make it work for `Fixtures`. (the etcd bit is in #161, and the apiserver bit is in #162).

More specifically:

`&Fixtures{}.Start()` should behave like `NewFixtures("some/sane/etcd/binary/path", "some/sane/apiserver/path").Start()`

totherme:
[closes #163]

In the process of making the APIServer constructor take 0 arguments (#162), we naturally find ourselves making the Fixtures constructor take 0 arguments (#163).

We'll be cleaning things up even more as we continue to work on #162, but this seems like a reasonable mid-point where there's already some user-value. In particular, the main interface to the test framework (The Fixtures constructor and methods) ought to be stable now:

```go
myFixtures, err := test.NewFixtures()
err = myFixtures.Start()
myServerURL = myFixtures.APIServerURL()
// Run tests using myServerURL
err = myFixtures.Stop()
```

Cc @hoegaarden @apelisse

### DIFF
--- a/pkg/framework/test/apiserver.go
+++ b/pkg/framework/test/apiserver.go
@@ -33,34 +33,18 @@ type certDirManager interface {
 
 var apiServerBinPathFinder = DefaultBinPathFinder
 
-// NewAPIServer constructs a new APIServer with whatever api_server binary it can find.
+// NewAPIServer creates a new APIServer Fixture Process
 func NewAPIServer(config *APIServerConfig) *APIServer {
 	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
 		return gexec.Start(command, out, err)
 	}
 
 	return &APIServer{
-		Path:           apiServerBinPathFinder("kube_apiserver"),
+		Path:           apiServerBinPathFinder("kube-apiserver"),
 		Config:         config,
-		ProcessStarter: starter,
-		CertDirManager: &TempDirManager{},
-	}
-}
-
-// NewAPIServerWithBinary creates a new APIServer Fixture Process
-func NewAPIServerWithBinary(pathToAPIServer string, config *APIServerConfig) *APIServer {
-	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
-		return gexec.Start(command, out, err)
-	}
-
-	apiserver := &APIServer{
-		Path:           pathToAPIServer,
 		ProcessStarter: starter,
 		CertDirManager: NewTempDirManager(),
-		Config:         config,
 	}
-
-	return apiserver
 }
 
 // Start starts the apiserver, waits for it to come up, and returns an error, if occoured.
--- a/pkg/framework/test/apiserver.go
+++ b/pkg/framework/test/apiserver.go
@@ -19,6 +19,7 @@ type APIServer struct {
 	ProcessStarter simpleSessionStarter
 	CertDirManager certDirManager
 	Config         *APIServerConfig
+	Etcd           FixtureProcess
 	session        SimpleSession
 	stdOut         *gbytes.Buffer
 	stdErr         *gbytes.Buffer
@@ -34,9 +35,14 @@ type certDirManager interface {
 var apiServerBinPathFinder = DefaultBinPathFinder
 
 // NewAPIServer creates a new APIServer Fixture Process
-func NewAPIServer(config *APIServerConfig) *APIServer {
+func NewAPIServer(config *APIServerConfig) (*APIServer, error) {
 	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
 		return gexec.Start(command, out, err)
+	}
+
+	etcd, err := NewEtcd()
+	if err != nil {
+		return nil, err
 	}
 
 	return &APIServer{
@@ -44,12 +50,23 @@ func NewAPIServer(config *APIServerConfig) *APIServer {
 		Config:         config,
 		ProcessStarter: starter,
 		CertDirManager: NewTempDirManager(),
-	}
+		Etcd:           etcd,
+	}, nil
+}
+
+// GetURL returns the URL APIServer is listening on. Clients can use this to connect to APIServer.
+func (s *APIServer) GetURL() string {
+	return s.Config.APIServerURL
 }
 
 // Start starts the apiserver, waits for it to come up, and returns an error, if occoured.
 func (s *APIServer) Start() error {
 	if err := s.Config.Validate(); err != nil {
+		return err
+	}
+
+	err := s.Etcd.Start()
+	if err != nil {
 		return err
 	}
 
@@ -74,7 +91,7 @@ func (s *APIServer) Start() error {
 		"--admission-control-config-file=",
 		"--bind-address=0.0.0.0",
 		"--storage-backend=etcd3",
-		fmt.Sprintf("--etcd-servers=%s", s.Config.EtcdURL),
+		fmt.Sprintf("--etcd-servers=%s", s.Etcd.GetURL()),
 		fmt.Sprintf("--cert-dir=%s", certDir),
 		fmt.Sprintf("--insecure-port=%s", clientURL.Port()),
 		fmt.Sprintf("--insecure-bind-address=%s", clientURL.Hostname()),
@@ -102,6 +119,7 @@ func (s *APIServer) Stop() {
 	if s.session != nil {
 		s.session.Terminate()
 		s.session.Wait(20 * time.Second)
+		s.Etcd.Stop()
 		err := s.CertDirManager.Destroy()
 		gomega.Expect(err).NotTo(gomega.HaveOccurred())
 	}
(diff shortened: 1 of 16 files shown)


pkg/framework/test/bin_path_finder.go
```diff
@@ -25,7 +26,11 @@ type BinPathFinder func(symbolicName string) (binPath string)
 // variable, derived from the symbolic name, and falls back to a default assets location when
 // this variable is not set
 func DefaultBinPathFinder(symbolicName string) (binPath string) {
-	envVar := "TEST_ASSET_" + strings.ToUpper(symbolicName)
+	punctuationPattern := regexp.MustCompile("[^A-Z0-9]+")
```
apelisse: I'd maybe rather have the code crash if the name is not valid? Rather than try to guess and do something that might be unexpected?

pkg/framework/test/apiserver.go
```diff
@@ -34,17 +35,28 @@ type certDirManager interface {
 var apiServerBinPathFinder = DefaultBinPathFinder
 
 // NewAPIServer creates a new APIServer Fixture Process
-func NewAPIServer(config *APIServerConfig) *APIServer {
+func NewAPIServer(config *APIServerConfig) (*APIServer, error) {
 	starter := func(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
 		return gexec.Start(command, out, err)
 	}
 
+	etcd, err := NewEtcd()
```
apelisse: Can we take this as a parameter instead? Or put that as part of the config and default to NewEtcd if unset?

pkg/framework/test/apiserver.go
```diff
@@ -53,6 +65,11 @@ func (s *APIServer) Start() error {
 		return err
 	}
 
+	err := s.Etcd.Start()
```
apelisse: I've been wondering about that for a while ... Should we expect the etcd process to be started or not? While it doesn't hurt to start if not started, I'm more curious about the parallel with stop: should we stop the etcd server when we stop the apiserver?

pkg/framework/test/apiserver.go
```diff
@@ -102,6 +119,7 @@ func (s *APIServer) Stop() {
 	if s.session != nil {
 		s.session.Terminate()
 		s.session.Wait(20 * time.Second)
+		s.Etcd.Stop()
```
apelisse: See, I don't know about that

pkg/framework/test/apiserver_config.go
```diff
@@ -4,7 +4,7 @@ import "github.com/asaskevich/govalidator"
 
 // APIServerConfig is a struct holding data to configure the API Server process
 type APIServerConfig struct {
-	EtcdURL      string `valid:"required,url"`
+	EtcdURL      string `valid:"url"`
```
apelisse: Can this be the Etcd object instead?

apelisse: Also, is this used? Since you connect to `etcd.GetURL` rather than this value now?

pkg/framework/test/apiserver_test.go
```diff
@@ -21,11 +21,13 @@ var _ = Describe("Apiserver", func() {
...
+		fakeEtcdProcess    *testfakes.FakeFixtureProcess
 	)
 
 	BeforeEach(func() {
 		fakeSession = &testfakes.FakeSimpleSession{}
 		fakeCertDirManager = &testfakes.FakeCertDirManager{}
+		fakeEtcdProcess = &testfakes.FakeFixtureProcess{}
 
 		apiServerConfig = &APIServerConfig{
 			EtcdURL:      "http://this.is.etcd:2345/",
```
apelisse: is this still used?

pkg/framework/test/fixtures.go
```diff
@@ -85,7 +82,6 @@ func (f *Fixtures) Start() error {
 // Stop will stop all your fixtures, and clean up their data.
 func (f *Fixtures) Stop() error {
 	f.APIServer.Stop()
-	f.Etcd.Stop()
```
apelisse: Then we could just keep that

pkg/framework/test/apiserver_config.go
```diff
@@ -4,7 +4,6 @@ import "github.com/asaskevich/govalidator"
 
 // APIServerConfig is a struct holding data to configure the API Server process
 type APIServerConfig struct {
-	EtcdURL      string `valid:"url"`
```
apelisse: OK 😂

pkg/framework/test/integration/integration_test.go
```diff
@@ -17,7 +17,7 @@ var _ = Describe("The Testing Framework", func() {
 		var err error
 		var fixtures *test.Fixtures
 
-		fixtures, err = test.NewFixtures(defaultPathToEtcd)
+		fixtures, err = test.NewFixtures()
```
apelisse: Isn't this hiding too much now? I can't configure the binaries, which means I can't test against multiple versions?  OK I can change the binaries in the env location, or change the env. 
Should we have a config that let's one change the default?

pkg/framework/test/apiserver.go
totherme: We're aiming to make these constructors take no params, so you can use them if you want sensible defaults. If you want to provide your own Etcd, you can either use `APIServer{Etcd: myEtcd}` or `myAPIServer := NewAPIServer() ; myAPIServer.Etcd = myEtcd`.

We're still part-way through removing the parameters of the APIServer constructor, but that's where we were aiming.

pkg/framework/test/apiserver.go
totherme: I think we've previously chatted about a couple of different potential down-the-line use-cases that drive this behaviour out. I'm now not convinced I understood everything we chatted about, so I'd like to repeat my understanding here to double-check. One key potential use case was the case in which the user wants to use an aggregate APIServer. This aggregate APIServer might be composed of a number of SubAPIServers, each of which needs to communicate with Etcd. At that time we believed we were comparing two possible architectures, which I will call "the old way" and "the new way".

In "the old way", every APIServer (and hence in future, presumably every SubAPIServer) contained a reference to some configuration which pointed to the address of an Etcd. This is enough to talk to an Etcd but not to start it. This meant that the starting Etcd was the job of the Fixtures struct. One major disadvantage of this approach was that if each potential SubAPIServer needed its own independent Etcd, this could not be arranged. The Fixtures struct only knows how to start one of them.

In "the new way", every APIServer (and hence in future, presumably every SubAPIServer) contains an actual Etcd struct. This means that it has the configuration needed to connect to the Etcd once it's running, and also that it has the means to start and stop it. In this world, starting and stopping Etcd becomes the job of the APIServer, not the Fixtures struct. This makes it possible for potential future SubAPIServers to manage their own state, without interfering with each other, and crucially without interference from the Fixtures struct. If you want to share a single Etcd between all your SubAPIServers, it can be the responsibility of the APIServerAggregator to start and stop Etcd. If you want different Etcds for different SubAPIServers, then each SubAPIServer can manage its own.

Were we talking at crossed-purposes when we last discussed this? Is it worth chatting about it again?

pkg/framework/test/bin_path_finder.go
totherme: This is because we're generating the environment variable from the executable name. So, if your asset-executable is called `my-asset` then you want your environment variable to be `TEST_ASSET_MY_ASSET`. The transformation is because typically you get different conventions for executable names (lower case, with dashes or underscores) than for environment variables (upper case, underscores only). In our case, we're working with an apiserver executable called `kube-apiserver`, and we want the configuration variable to be `TEST_ASSET_KUBE_APISERVER`. Bash would interpret `$TEST_ASSET_KUBE-APISERVER` as the environment variable `TEST_ASSET_KUBE` followed by the literal string `-APISERVER`.

The transformation strategy here owes something to the approach used by the spring framework, which allows you to set your configuration options using environment variables, build parameters, command line flags, and who knows what else -- all using the appropriate conventions for the medium. If this is too much magic for this context, we could certainly remove everything except the translation from `-` to `_`. Or we could insist that all test binaries have names which correspond to valid bash environment variables.

pkg/framework/test/integration/integration_test.go
totherme: You can configure the binaries by:
- Putting different binaries in the default path `assets/bin` (relative to the test framework package)
- Changing environment variables in your test environment (eg `TEST_ASSET_ETCD`)
- Changing the pathfinder in code:
```
fixtures = test.NewFixtures()
apiServer = test.NewAPIServer() // currently this takes more params, but we're working on it
apiServer.PathFinder = func(_ string) string {
  return "/my/path/to/apiserver"
}
fixtures.APIServer = apiServer
fixtures.Start()
```

PR #169, opened 2017-12-11
Move several packages

mengqiy:
k8s.io/kubernetes/pkg/kubectl/categories -> k8s.io/kubectl/pkg/categories
k8s.io/kubernetes/pkg/kubectl/scheme -> k8s.io/kubectl/pkg/scheme
k8s.io/kubernetes/pkg/kubectl/validation -> k8s.io/kubectl/pkg/validation


mengqiy:
resource pkg will be in a separate PR after this one is in.

dims:
@mengqiy shouldn't we be moving stuff to staging/src/k8s.io/kubectl So we can move the whole thing out down the line? (curious)

mengqiy:
@dims ICYM, https://docs.google.com/document/d/1kzRLeRZJolxvyl_zaGH9qYp3jUKowUxGqdrvlSmMCO8/edit?usp=sharing

dims:
@mengqiy thanks for the doc. looks like @sttts also mentioned the staging/ area in there and i don't see a response to that comment

apelisse:
I'll try to describe my thought-process so that one can know what I potentially missed:
1. I was wondering about all the merges, but yeah, you've needed one per package that you moved. Would it have been possible to create a pull-request for each? I'm assuming the dependency should create a DAG, so you should be able to topologically sort these and submit one PR for each :-) (I sound like a huge nit, sorry :D)
2. Let's `git diff --stat` to have a good look at what files you changed. Note for later, I don't think `pkg/validation/testdata/v1/invalidPod.yaml` are needed. Unrelated to this pull-request.
3. `> dep ensure --no-vendor -v: Gopkg.lock was already in sync with imports and Gopkg.toml`. Good.

Looks good to me. I'll leave it some time so that someone else can have an opinion.

sttts:
This PR is moving out packages depening on other `k8s.io/*` repos, with the plan to revendor them as k8s.io/kubectl into the main repo. This is a circular dependency between repos:

`k/k` -> `k8s.io/kubectl` -> `k8s.io/{apimachinery,client-go}` = `k/k`

We had a similar situation with `k8s.io/metrics` and had a very hard time with that: it stopped us to do changes in `k/k` that `k8s.io/metrics` depended on. I don't want to see the same situation again. I am against this PR.

Hence, **I want to put this approach on hold** until this is discussed deeply and decided in sig-architecture.

/cc @kubernetes/sig-architecture-pr-reviews @deads2k @smarterclayton

deads2k:
> This PR is moving out packages depening on other k8s.io/* repos, with the plan to revendor them as k8s.io/kubectl into the main repo. This is a circular dependency between repos:

I think the dependency problem is one that isn't obvious if you haven't walked it before.  If kubectl were ever vendored into kuberentes, then the level of apimachinery and client-go being used would suddenly jump to the kuberentes branch level, which may or may not work depending on how closely kubectl tracks.  We deal with it quite often in openshift and it is extremely painful.

While I'm in favor of the change, if you ever want to vendor into kubernetes (not sure if this is a goal or not), you'll want to live in staging like the rest.  Otherwise, kubectl will end up as debt that freezes the lower level libraries.

deads2k:
> We had a similar situation with k8s.io/metrics and had a very hard time with that: it stopped us doing changes in k/k that k8s.io/metrics depended on. I don't want to see the same situation again. I am against this PR.

Vendoring a library which vendors kubernetes isn't something we should do.  We fought this problem in the past with heapster and it took a few months to finally unwind.  You can end up in a state where k/k can't change because its vendor directory is pinning its APIs and the other project (heapster) cannot change because it had no new APIs to vendor.  You end up stepping through each package dependencies in order, duplicating apis in k/k, then vendoring into heapster, then revendoring into k/k, then deleting an original in k/k, and moving to the next package.  That isn't a tenable thing to do and it is exactly the reason that nothing ever moves into staging without having its links snipped *before* the move.

apelisse:
> it stopped us doing changes in k/k that k8s.io/metrics depended on.

If this is what you meant, yeah, something is very wrong ... How can we have `k8s.io/metrics` depend on kubernetes? Shouldn't it be the other way around?

mengqiy:
@sttts is right. Moving these packages may not work well.

So change of plan. After discussing with @apelisse @droot, now we think forking is the approach we want to take, which is described in https://docs.google.com/document/d/1yba5jRJSVf1IIXYBO11N8nHbwTbzkDFDssADIC4nbig/edit?usp=sharing.

mengqiy:
We still need this PR in even we do forking.

mengqiy:
Closing this since this won't work because of the `/staging`

PR #173, opened 2017-12-12
Move openapi

apelisse:
Copy the openapi package from kubernetes `k8s.io/kubernetes/pkg/kubectl/cmd/util/openapi` to `k8s.io/kubectl/pkg/openapi/`.

@AlexKhotian has done pretty much everything. I've just simplified some of the history.


mengqiy:
@apelisse Phil suggests moving all the content in this PR to a new repo (maybe called CLI-lib) under [kubernetes-SIGs](https://github.com/kubernetes-SIGs) org. So kinflate and kubectl can vendor it. 
Thoughts?

apelisse:
As discussed, we have a plan sort-of. We'll drop this pull-request (big thank you to @AlexKhotian for doing it, sorry it's not going anywhere :-/), Ellen already copied the openapi code because we needed to to move forward, so we have that in the mean time. And the final solution will probably be to move some of this client generic code into staging in kubernetes, so that it can be used from everywhere.

monopole:
@apelisse can this be closed?

apelisse:
Good point. This has been done actually ...

PR #176, opened 2017-12-14
162 apiserver sane default constructor

### LINKED ISSUE #162: In the testing framework, the APIServer struct should have sane 0-value behaviour

hoegaarden:
Based on #168:

The APIServer has now sane defaults for the constructor.
- The Port/Address allocation/management is handled by a (new thing called) AddressManager
- The path to APIServer's binary is now managed by a PathFinder

[closes #162]

@totherme


hoegaarden:
/assign @apelisse 

Note: This PR is based on #168, so it looks way bigger then it actually is.

PR #177, opened 2017-12-14
Zero conf structs

hoegaarden:
Based on: #176 

This is some more cleanup and work on the test framework. The most important changes of this PR are:
- Rename `Fixtures` to `ControlPlane` (and `FixtureProcess` to `ControlPlaneProcess`, and others like that)
- Have `Etcd` & `APIServer` work as 0-conf structs, just use `etcd := &Etcd{}`
- Remove `PortFinder` in favour of `AddressManager`
- Remove both `EtcdConfig` and `APIServerConfig`
- Some additional testing

@totherme


hoegaarden:
/assign @apelisse 

Note: This PR is based on #176, so it looks way bigger then it actually is.

pkg/framework/test/apiserver.go
```diff
@@ -97,14 +103,47 @@ func (s *APIServer) Start() error {
...
+	s.session.Terminate()
+	// TODO have a better way to handle the timeout of Stop()
+	s.session.Wait(20 * time.Second)
+
+	err := s.Etcd.Stop()
+	if err != nil {
+		return err
 	}
+
+	err = s.CertDirManager.Destroy()
```
apelisse: nit: why save error and then return it?

apelisse: Also, shouldn't we try to remove the certdirmanager even if we failed to stop?

apelisse: Same for etcd of course (which reminds me that there is so much in common ...)

pkg/framework/test/control_plane.go
```diff
@@ -0,0 +1,41 @@
...
+//
+// Right now, that means Etcd and your APIServer. This is likely to increase in future.
+type ControlPlane struct {
+	APIServer ControlPlaneProcess
+}
+
+// ControlPlaneProcess knows how to start and stop a ControlPlane process.
+// This interface is potentially going to be expanded to e.g. allow access to the processes StdOut/StdErr
+// and other internals.
+type ControlPlaneProcess interface {
```
apelisse: I don't know if the word "Process" is fair, since it's not technically a process, but a bunch of processes.

pkg/framework/test/apiserver.go
hoegaarden: What to do when the `Stop()` fails?
My first instinct was: We have no idea in which state the system (the APIServer in that case) is. We don't know, if it is still connected to Etcd, if it has files open in de cert dir, ... so for now we just bail out and let the things in the state they are.
In future we can certainly think about if and how we can be smarter about that, if we can recover from certain errors and retry, ... and all those sort of things.

PR #180, opened 2017-12-15
Examples

hoegaarden:
- Refactor AddressManager
- Add documentation for APIServer (and Etcd) and add some examples

Cc @totherme 

/assign @apelisse

### DIFF
--- a/pkg/framework/test/examples_api_server_custom_pathfinder_test.go
+++ b/pkg/framework/test/examples_api_server_custom_pathfinder_test.go
@@ -1,0 +1,19 @@
+package test_test
+
+import (
+	"fmt"
+
+	. "k8s.io/kubectl/pkg/framework/test"
+)
+
+func ExampleAPIServer_specialPathFinder() {
+	apiServer := &APIServer{
+		PathFinder: myCustomPathFinder,
+	}
+
+	fmt.Println(apiServer)
+}
+
+func myCustomPathFinder(_ string) string {
+	return "/usr/local/bin/kube/1.21-alpha1/special-kube-api"
+}
--- a/pkg/framework/test/examples_api_server_forced_args_process_starter_test.go
+++ b/pkg/framework/test/examples_api_server_forced_args_process_starter_test.go
@@ -1,0 +1,32 @@
+package test_test
+
+import (
+	"io"
+	"os/exec"
+
+	"fmt"
+
+	"github.com/onsi/gomega/gexec"
+	. "k8s.io/kubectl/pkg/framework/test"
+)
+
+func ExampleAPIServer_forcedArgsProcessStarter() {
+	apiServer := &APIServer{
+		ProcessStarter: forceArgsStarter,
+	}
+	fmt.Println(apiServer)
+}
+
+func forceArgsStarter(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
+	forcedArgs := []string{
+		"--target-ram-mb=1024",
+		"--allow-privileged=false",
+		"--authorization-mode=RBAC,AlwaysDeny",
+	}
+
+	newArgs := append(command.Args, forcedArgs...)
+
+	command.Args = newArgs
+
+	return gexec.Start(command, out, err)
+}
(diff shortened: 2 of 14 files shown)


pkg/framework/test/examples_api_server_custom_pathfinder_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+
+	. "k8s.io/kubectl/pkg/framework/test"
+)
+
+func ExampleAPIServer_specialPathFinder() {
+	apiServer := &APIServer{
+		PathFinder: myCustomPathFinder,
```
apelisse: I still don't understand why this is more complicated than it has to be?

My understanding is that you want lazy-finding of the binary. Do you assume that the decision can't be made at APIServer *creation* time, but can be made at APIServer *start* time? Why would we expect that to be different? If we don't, we can just let the user compose however they want. That would also create a more gradual learning curve (*I don't need to know new types if I just want to pass in an absolute path*, *I can understand how the APIServer works by looking just at its interface*)

pkg/framework/test/examples_api_server_custom_pathfinder_test.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+
+	fmt.Println(apiServer)
```
apelisse: Since you don't have a `// Output:` line, this code will not run with tests, not sure this line is useful.

pkg/framework/test/examples_api_server_custom_pathfinder_test.go
apelisse: Actually, this example wouldn't be needed at all then.

pkg/framework/test/examples_api_server_forced_args_process_starter_test.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+func ExampleAPIServer_forcedArgsProcessStarter() {
+	apiServer := &APIServer{
+		ProcessStarter: forceArgsStarter,
+	}
+	fmt.Println(apiServer)
+}
+
+func forceArgsStarter(command *exec.Cmd, out, err io.Writer) (SimpleSession, error) {
```
apelisse: That seems like a valid feature, but I have the feeling that:
- Asking users to run `gexec.Start` might be too much,
- Exposing the `SimpleSession` shouldn't be necessary

Beside arguments, do we expect them to change anything else?

pkg/framework/test/examples_api_server_special_cert_dir_manager_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"io/ioutil"
+	"os"
+
+	. "k8s.io/kubectl/pkg/framework/test"
+)
+
+func ExampleAPIServer_credHubCertDirManager() {
+	apiServer := &APIServer{
+		CertDirManager: NewCredHubCertDirManager(),
```
apelisse: Again, can't we just provide a path? If the path is present, we assume it's user managed. If the path is not, then it's apiserver managed. Then we can hide a *ton* of the directory management thing (much simpler to learn). If I provide my directory, I don't think I want the apiserver object to mess up with it too much (maybe I want to re-use it). And I shouldn't have to create a new class for that, or learn about "TempDirManager", "Remover" and all.

apelisse: And this example then becomes unnecessary

pkg/framework/test/temp_dir_manager.go
```diff
@@ -6,9 +6,13 @@ import (
 )
 
 // TempDirMaker can create directories.
+//
+// It is implemented by `ioutil.TempDir`
 type TempDirMaker func(dir, prefix string) (name string, err error)
 
 // TempDirRemover can delete directories
+//
+// It is implemented by `os.RemoveAll`
```
apelisse: Same

pkg/framework/test/temp_dir_manager.go
```diff
@@ -6,9 +6,13 @@ import (
 )
 
 // TempDirMaker can create directories.
+//
+// It is implemented by `ioutil.TempDir`
```
apelisse: I don't want to have to learn about that :-/

pkg/framework/test/examples_api_server_custom_pathfinder_test.go
totherme: Agreed -- addressed in #184

PR #183, opened 2017-12-15
update vendor

mengqiy:
Pin the following repo to master branch:
- api
- apimachinery
- client-go


mengqiy:
Pick up some client-go change, e.g. https://github.com/kubernetes/kubectl/pull/183/files#diff-b2e852a96204fb6d4cce1c38965504f8 which is used by https://github.com/kubernetes/kubectl/blob/master/pkg/scheme/scheme.go#L32.

apelisse:
What process did you use to get the update?

mengqiy:
dep ensure

PR #193, opened 2018-01-05
Simplify directory handling

hoegaarden:
We now use simple strings for the paths to the certificate directory (APIServer) and the data directory (Etcd) instead of the `TempDirManager`.

cc: @totherme @apelisse


apelisse:
as discussed offline, I'm not a big fan of the directory, let's iterate on that.

PR #205, opened 2018-01-10
more support for configmap and secret

mengqiy:
There will be some test to replace https://github.com/kubernetes/kubectl/blob/c61d60e246663574dd531a0c61a99c325e81f809/cmd/kinflate/kinflate_test.go#L27-L32 soon.


mengqiy:
Test added.

PR #207, opened 2018-01-15
Test framework: Simplify UX

totherme:
In order to simplify the UX of our library, we decided to temporarily abandon our precious unit tests, and explore driving out the library interface using only the godocs. These commits tell that story.

The end result is that most of the actual logic has been moved into an internal package, and the visible parts of the library are largely configuration structs.

CC: @hoegaarden 
/assign @apelisse


hoegaarden:
This PR has been rebased onto https://github.com/kubernetes/kubectl/pull/208, do please review/merge that first.

PR #209, opened 2018-01-16
Test framework: Configure the APIServer with the URL to Etcd

hoegaarden:
The `APIServer` does not create, start & stop `Etcd` itself anymore, but gets configured with the URL to `Etcd`. Creating, starting & stopping of `Etcd` is now the responsibility of the `ControlPlane`.

This PR is based on #207, thus it looks way bigger then it actually is.

CC: @totherme 
/assign @apelisse


trestronni:
@trestronni Approved this pull

PR #213, opened 2018-01-17
Test framework: Update docs

hoegaarden:
This is based on #209, so agian appears to be much bigger than it actually is.

CC: @totherme 
/assign @apelisse


pkg/framework/test/apiserver.go
```diff
@@ -10,35 +10,42 @@ import (
...
 	// If this is not specified, we default to a random free port on localhost.
 	URL *url.URL
 
-	// Path is the path to the apiserver binary. If this is left as the empty
-	// string, we will attempt to locate a binary, by checking for the
-	// TEST_ASSET_KUBE_APISERVER environment variable, and the default test
-	// assets directory.
+	// Path is the path to the apiserver binary.
+	//
+	// If this is left as the empty string, we will attempt to locate a binary,
```
apelisse: I'm not sure I understand what's going to happen based on that sentence.

is `TEST_ASSET_KUBE_APISERVER` the name of the binary? I think so since it has the APISERVER name in it.
But then,

> the default test assets directory

what binary do you take there?

pkg/framework/test/apiserver.go
```diff
@@ -10,35 +10,42 @@ import (
...
+	// by checking for the TEST_ASSET_KUBE_APISERVER environment variable, and
+	// the default test assets directory.
 	Path string
 
-	// CertDir is a struct holding a path to a certificate directory and a function to cleanup that directory.
+	// CertDir is a path to a directory containing whatever certificates the
+	// APIServer will need.
+	//
+	// If left unspecified, then the Start() method will create a fresh temporary
+	// directory, and the Stop() method will clean it up.
```
apelisse: :+1:

pkg/framework/test/doc.go
```diff
@@ -0,0 +1,52 @@
...
+Etcd: Manages an Etcd binary, which can be started, stopped and connected to.
+Etcd will listen on a random port for http connections and will create a
+temporary directory for it'd data; unless configured differently.
+
+APIServer: Manages an Kube-APIServer binary, which can be started, stopped and
+connected to. APIServer will listen on a random port for http connections and
+will create a temporary directory to store the (auto-generated) certificates;
+unless configured differently.
+
+Binaries
```
apelisse: Oh ok, that helps :-)

pkg/framework/test/doc.go
```diff
@@ -0,0 +1,52 @@
+/*
+
+Package test implements a integration testing framework for kubernetes.
+
+It provides a kubernetes API you can connect to and test your
+kubernetes client implementations against. The lifecycle of the components
```
apelisse: I'd really like if we could not phrase this framework as a "client test framework". One could want to test a new etcd version by switching the binary. One could decide to test the apiserver, an extension apiserver, an aggregator, a FooBar component somewhere, etc. I think these are built as reusable pieces that can be composed to test any component in a kubernetes cluster.

pkg/framework/test/doc.go
```diff
@@ -0,0 +1,52 @@
...
+connected to. APIServer will listen on a random port for http connections and
+will create a temporary directory to store the (auto-generated) certificates;
+unless configured differently.
+
+Binaries
+
+Both Etcd and APIServer use the same mechanism to determine which binaries to
+use when they get started.
+
+1. If the component is configured with a `Path` the framework tries to run that
```
apelisse: We might want later to make the "controlplane" configurable with the same kind of defaulting mechanisms that we've used for APIServer and Etcd. But yeah, that's fine for now.

pkg/framework/test/doc.go
```diff
@@ -0,0 +1,52 @@
...
+binary. (Note: To overwrite the `Path` of a component you cannot use the
+ControlPlane, but you need to create, wire and start & stop all the components
+yourself.)
+
+2. If a environment variable named
+`TEST_ASSET_KUBE_APISERVER` or `TEST_ASSET_ETCD` is set, this value is used as a
+path to the binary for the APIServer or Etcd.
+
+3. If neither an environment variable is set nor the `Path` is overwritten
+explicitly, the framework tries to use the binaries apiserver or etcd in the
```
apelisse: If these are the actual names that you're looking for, then maybe you should quote them somehow.

apelisse:
Looks great!

https://godoc.org/k8s.io/kubectl/pkg/framework/test

PR #214, opened 2018-01-17
update mutate func

mengqiy:
update mutate func:
- mutateField now takes a list of mutation func
- support create if not present

Fix a bug in the last commit.

### DIFF
--- a/pkg/kinflate/update.go
+++ b/pkg/kinflate/update.go
@@ -30,29 +30,37 @@ import (
 	ptvisitor "k8s.io/kubectl/pkg/kinflate/pod_template_visitor"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/kinflate/update.go
```diff
@@ -30,37 +30,47 @@ import (
...
 	}
 
 	v, found := m[pathToField[0]]
 	if !found {
-		return nil
+		if createIfNotPresent {
+			v = map[string]interface{}{}
+			m[pathToField[0]] = v
+		} else {
+			return nil
```
monopole: Please make this an early return, e.g. 
```
if !found {
  if !createIfNotPresent {
    return nil
  }
  m[pathToField[0]] = map[string]interface{}{}
}
```

trestronni:
I need help. I keep getting 404 error pages??

PR #215, opened 2018-01-19
Nameprefix util

mengqiy:
Add
- encode and decode util
- nameprefix util


pkg/kinflate/util/gvkn_sort.go
```diff
@@ -0,0 +1,28 @@
+/*
```
mengqiy: This file is copied from `pkg/kinflate/gvkn_sort.go`
The original copy will be deleted after `kinflate` switched to use `pkg/kinflate/util`.

pkg/kinflate/util/prefixname.go
```diff
@@ -0,0 +1,83 @@
...
+
+	err = o.Run(m)
+	if err != nil {
+		return nil, err
+	}
+
+	return Encode(m)
+}
+
+func (o *PrefixNameOptions) addPrefix(in interface{}) (interface{}, error) {
```
mengqiy: This func is copied from `https://github.com/kubernetes/kubectl/blob/master/pkg/kinflate/update.go#L100`

pkg/kinflate/util/util.go
```diff
@@ -0,0 +1,161 @@
...
+			return nil, err
+		}
+		firstObj = false
+	}
+	return buf.Bytes(), nil
+}
+
+type mutateFunc func(interface{}) (interface{}, error)
+
+func mutateField(m map[string]interface{}, pathToField []string, createIfNotPresent bool, fns ...mutateFunc) error {
```
mengqiy: This func is copied from https://github.com/kubernetes/kubectl/blob/master/pkg/kinflate/update.go#L35

pkg/kinflate/util/prefixname.go
```diff
@@ -0,0 +1,83 @@
...
+		pathsConfig = DefaultNamePrefixPathsConfig
+	}
+	o.pathsConfig = pathsConfig
+}
+
+func (o *PrefixNameOptions) Transform(m map[GroupVersionKindName]*unstructured.Unstructured) error {
+	for gvkn := range m {
+		obj := m[gvkn]
+		for _, path := range o.pathsConfig.Paths {
+			objMap := obj.UnstructuredContent()
```
mengqiy: I will push the change to move this line out of the for loop.

pkg/kinflate/util/interface.go
```diff
@@ -0,0 +1,33 @@
...
+	// Transform takes a map GroupVersionKindName->Unstructured and do
+	// some transformation (e.g. prefix name or apply additional labels) on
+	// the objects in the same map.
+	Transform(m map[GroupVersionKindName]*unstructured.Unstructured) error
+	// TransformBytes is a wrapper of func Transform.
+	// It takes a byte array, converts it into a map GroupVersionKindName->Unstructured.
+	// Then call Transform to do the transformation.
+	// It returns a byte array of encoded transformed objects.
+	TransformBytes([]byte) ([]byte, error)
+}
```
monopole: this file should be called transformer.go

An interface cannot commit itself to be a wrapper of another interface or to be implemented in a particular way.

Suggest one interface:
```
type Transformer interface {
  // Transform modifies objects in a map, e.g. add prefixes or additional labels.
  Transform(m map[GroupVersionKindName]*unstructured.Unstructured) error
}
``

because this interface:
```
type Transformer interface {
  // Transform converts an array of bytes into a new array of bytes.
  Transform([]byte) ([]byte, error)
}
```
is too general - it describes every program ever written.  

Besides this seems to be going down a road where there are many different implementations of the GVKM map transformer, but only one implementation of the []byte to []byte transformer (which will marshal, call transform, then unmarshal)

File: pkg/kinflate/util/pathconfig.go [hunk 6]
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package util
+
+type PathConfig struct {
+	Path               []string
+	CreateIfNotPresent bool
+}
+
+type PathsConfig struct {
```
monopole: PathConfigs, not PathsConfig

i'm not convinced that using this type is clearer than using []PathConfig directly.  are you?

pkg/kinflate/util/prefixname.go
```diff
@@ -0,0 +1,83 @@
...
+
+package util
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+type PrefixNameOptions struct {
```
monopole: This needs godoc

PR #218, opened 2018-01-22
additional labels and annotations util

mengqiy:
An implementation of `Transformer` interface to apply additional labels annotations.

### DIFF
--- a/pkg/kinflate/util/labelsandannotations.go
+++ b/pkg/kinflate/util/labelsandannotations.go
@@ -17,41 +17,45 @@ limitations under the License.
 package util
 
 import (
+	"errors"
 	"fmt"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-// MapTransformationOptions contains a map string->string and path configs
+// MapTransformer contains a map string->string and path configs
 // The map will be applied to the fields specified in path configs.
-type MapTransformationOptions struct {
+type MapTransformer struct {
 	m           map[string]string
 	pathConfigs []PathConfig
 }
 
-var _ Transformer = &MapTransformationOptions{}
+var _ Transformer = &MapTransformer{}
 
-// CompleteForLabels fills up the MapTransformationOptions for labels transformation.
-func (o *MapTransformationOptions) CompleteForLabels(m map[string]string, pathConfigs []PathConfig) {
-	o.m = m
-	if pathConfigs == nil {
-		pathConfigs = DefaultLabelsPathConfigs
-	}
-	o.pathConfigs = pathConfigs
+// NewDefaultingLabelsMapTransformer construct a MapTransformer with defaultLabelsPathConfigs.
+func NewDefaultingLabelsMapTransformer(m map[string]string) (Transformer, error) {
+	return NewMapTransformer(defaultLabelsPathConfigs, m)
 }
 
-// CompleteForAnnotations fills up the MapTransformationOptions for annotations transformation.
-func (o *MapTransformationOptions) CompleteForAnnotations(m map[string]string, pathConfigs []PathConfig) {
-	o.m = m
-	if pathConfigs == nil {
-		pathConfigs = DefaultAnnotationsPathConfigs
-	}
-	o.pathConfigs = pathConfigs
+// NewDefaultingAnnotationsMapTransformer construct a MapTransformer with defaultAnnotationsPathConfigs.
+func NewDefaultingAnnotationsMapTransformer(m map[string]string) (Transformer, error) {
+	return NewMapTransformer(defaultAnnotationsPathConfigs, m)
 }
 
-// Transform apply each <key, value> pair in the MapTransformationOptions to the
-// fields specified in MapTransformationOptions.
-func (o *MapTransformationOptions) Transform(m map[GroupVersionKindName]*unstructured.Unstructured) error {
+// NewMapTransformer construct a MapTransformer.
+func NewMapTransformer(pc []PathConfig, m map[string]string) (Transformer, error) {
+	if m == nil {
+		return nil, errors.New("map is not expected to be nil")
+	}
+	if pc == nil {
+		return nil, errors.New("pathConfigs is not expected to be nil")
+	}
+	return &MapTransformer{pathConfigs: pc, m: m}, nil
+}
+
+// Transform apply each <key, value> pair in the MapTransformer to the
+// fields specified in MapTransformer.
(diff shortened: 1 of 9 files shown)


mengqiy:
Non-test and non-boilerplate code is only about 150 lines :)

pkg/kinflate/util/labelsandannotations_test.go
```diff
@@ -0,0 +1,320 @@
...
+			name: "deploy1",
+		}: getDeployment(),
+		{
+			gvk:  schema.GroupVersionKind{Version: "v1", Kind: "Service"},
+			name: "svc1",
+		}: getService(),
+	}
+}
+
+var labeledObj1 = unstructured.Unstructured{
```
monopole: see commend above, thanks

pkg/kinflate/util/util.go
```diff
@@ -111,6 +112,53 @@ func Encode(in map[GroupVersionKindName]*unstructured.Unstructured) ([]byte, err
 	return buf.Bytes(), nil
 }
 
+// SelectByGVK returns true if selector is a superset of in; otherwise, false.
```
monopole: Please define superset with some examples.

its not obvious to me that if selector is nil, we have a superset condition.  is nil like a wildcard then?

pkg/kinflate/util/util.go
```diff
@@ -111,6 +112,53 @@ func Encode(in map[GroupVersionKindName]*unstructured.Unstructured) ([]byte, err
...
+	}
+	if len(selector.Kind) > 0 {
+		if in.Kind != selector.Kind {
+			return false
+		}
+	}
+	return true
+}
+
+func CompareMap(m1, m2 map[GroupVersionKindName]*unstructured.Unstructured) error {
```
monopole: If CompareMap is a test-only utility it should be in a _test.go file and should be private.

If it is meant to be used outside tests, then it needs its own individual test and godoc.

mengqiy:
@monopole Addressed the comments.
PTAL at https://github.com/kubernetes/kubectl/pull/218/files#diff-c1f60fb8ec4a94908d114d943a2fa297R27 to check if the name is good enough :)

pkg/kinflate/util/labelsandannotations.go
```diff
@@ -22,35 +22,41 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-type ApplyAdditionalMapOptions struct {
-	additionalMap map[string]string
-	pathConfigs   []PathConfig
+// MapTransformationOptions contains a map string->string and path configs
+// The map will be applied to the fields specified in path configs.
+type MapTransformationOptions struct {
```
monopole: Please call this MapTransformer, or TransformerByMap, or something.  

Its an implementation of Transformer.

pkg/kinflate/util/labelsandannotations.go
```diff
@@ -22,35 +22,41 @@ import (
...
 	if pathConfigs == nil {
 		pathConfigs = DefaultLabelsPathConfigs
 	}
 	o.pathConfigs = pathConfigs
 }
 
-func (o *ApplyAdditionalMapOptions) CompleteForAnnotations(m map[string]string, pathConfigs []PathConfig) {
-	o.additionalMap = m
+// CompleteForAnnotations fills up the MapTransformationOptions for annotations transformation.
+func (o *MapTransformationOptions) CompleteForAnnotations(m map[string]string, pathConfigs []PathConfig) {
```
monopole: likewise drop

pkg/kinflate/util/labelsandannotationsconfig.go
```diff
@@ -20,6 +20,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
+// DefaultLabelsPathConfigs is the default configuration for mutating labels and
+// selector fields for native k8s APIs.
 var DefaultLabelsPathConfigs = []PathConfig{
```
monopole: It seems to me these can be package private.  They could even be embedded in the constructors suggested above.

pkg/kinflate/util/util.go
```diff
@@ -112,7 +119,13 @@ func Encode(in map[GroupVersionKindName]*unstructured.Unstructured) ([]byte, err
...
 
-// SelectByGVK returns true if selector is a superset of in; otherwise, false.
+// SelectByGVK returns true if `selector` selects `in`; otherwise, false.
+// If `selector` and `in` are the same, return true.
+// If `selector` is nil, it is considered as a wildcard and always return true.
+// e.g. selector <Group: "", Version: "", Kind: "Deployemt"> CAN select
+// <Group: "extensions", Version: "v1beta1", Kind: "Deployemt">.
+// selector <Group: "apps", Version: "", Kind: "Deployemt"> CANNOT select
+// <Group: "extensions", Version: "v1beta1", Kind: "Deployemt">.
 func SelectByGVK(in schema.GroupVersionKind, selector *schema.GroupVersionKind) bool {
```
monopole: thanks much better!

pkg/kinflate/util/labelsandannotations.go
```diff
@@ -0,0 +1,80 @@
...
+// The map will be applied to the fields specified in path configs.
+type MapTransformationOptions struct {
+	m           map[string]string
+	pathConfigs []PathConfig
+}
+
+var _ Transformer = &MapTransformationOptions{}
+
+// CompleteForLabels fills up the MapTransformationOptions for labels transformation.
+func (o *MapTransformationOptions) CompleteForLabels(m map[string]string, pathConfigs []PathConfig) {
```
monopole: sorry, previous review somehow dropped a comment here.

This is basically a double setter with a hidden default, and raises alarms for me.

Please drop this, in favor of an object that is immutable behind the Transformer interface:

```
func NewDefaultingLabelsMapTransformer(m map[string]string) (Transformer, err) {
  return NewMapTransformer(DefaultLabelsPathConfigs, m)
}

func NewDefaultingAnnotationsMapTransformer(m map[string]string) (Transformer, err) {
  return NewMapTransformer(DefaultAnnotationsPathConfigs, m)
}

func NewMapTransformer(pc []PathConfig, m map[string]string) (Transformer, err) {
  if (m == nil or is otherwise bad in some way) {
   return nil, errors.New("bad map")
  }
  if (pc == nil or is otherwise bad in some way) {
   return nil, errors.New("bad pathconfigs")
  }
  return &MapTransformer{pc, m}, nil
}
```

mengqiy:
Updated. Will squash if you want me to do so.

PR #225, opened 2018-01-24
Rebase kinflate on new utils

mengqiy:
Rebase kinflate on new utils and remove some dead code.
kinflate no longer requires overlays. i.e. each manifest can be applied alone.
This PR addresses the 1st issue in https://github.com/kubernetes/kubectl/pull/220#issue-290884591


pkg/kinflate/adjust_path.go
```diff
@@ -17,18 +17,17 @@ limitations under the License.
...
 
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
 )
 
-func adjustPathsForConfigMapAndSecret(in *manifest.Manifest, pathToDir []string) *resource {
-	out := &resource{}
-	out.configmaps = adjustPathForConfigMaps(in.Configmaps, pathToDir)
-	out.secrets = adjustPathForSecrets(in.Secrets, pathToDir)
-	return out
+func adjustPathsForManifest(manifest *manifest.Manifest, pathToDir []string) {
```
monopole: s/manifest/m/

pkg/kinflate/configmap_secret.go
```diff
@@ -18,74 +18,114 @@ package kinflate
...
-		err = populateMap(m, corev1secret, nameWithHash)
+		err = populateMap(m, unstructuredSecret, nameWithHash)
 		if err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func constructConfigmapAndGenerateName(cm manifest.ConfigMap) (*unstructured.Unstructured, string, error) {
```
monopole: s/construct/make/

here and below.   because make is shorter and more "go"-ish

pkg/kinflate/util.go
```diff
@@ -17,122 +17,232 @@ limitations under the License.
...
+			return nil, err
+		}
 	}
+	return into, nil
+}
 
-	if len(overlay.Resources) == 0 {
-		return nil, nil, nil, errors.New("expect at least one base, but got 0")
+// DirToMap tries to find Kube-manifest.yaml first in a dir.
+// If not found, traverse all the file in the dir.
```
monopole: it does more than that - it does recursively finds all instances of Kube-manifest, reads them and merges them all into "into".  

The public entry to this recursion should be a method that does not take "into".  Callers should not be aware of this private accumulator.

Even better, can we tease apart the recursive file traversal from the actual accumulation into a map[kutil.GroupVersionKindName]*unstructured.Unstructured ?
  
Would make it easier to read and test.  The file traversal could return a channel of struct {*Manifest, error} for example.  then something could range over that channel.  the range ends when the channel closes.  or less fancy return an array of manifest, but a channel is actually easier in go.

pkg/kinflate/util.go
```diff
@@ -17,122 +17,232 @@ limitations under the License.
...
 		}
-		m[gvkn] = content
 	}
-	return nil
+	return baseResouceMap, nil
+}
+
+// DefaultTransformers generates 4 transformers:
+// 1) name prefix 2) apply labels 3) apply annotations 4) update name reference
+func DefaultTransformers(manifest *manifest.Manifest) ([]kutil.Transformer, error) {
```
monopole: private

pkg/kinflate/util.go
```diff
@@ -17,122 +17,232 @@ limitations under the License.
...
+	err = populateConfigMapAndSecretMap(manifest, baseResouceMap)
+	if err != nil {
+		return nil, err
+	}
+
+	transformers, err := DefaultTransformers(manifest)
+	if err != nil {
+		return nil, err
+	}
+	for _, transformer := range transformers {
```
monopole: s/transformer/t/  (it's a short loop)

BTW, love how this turned out.  Good effort.

pkg/kinflate/configmap_secret.go
```diff
@@ -18,74 +18,114 @@ package kinflate
...
+		return nil, err
+	}
+	var out unstructured.Unstructured
+	err = out.UnmarshalJSON(marshaled)
+	return &out, err
+}
+
+// We cannot marshal a corev1 Secret and then unmarshal it to a unstructured.
+// Because there are []byte fields (Data is map[string][]byte) in corev1 Secret.
+// In goland, unmarshal an marshaled byte array will become a base64 encoded string.
```
monopole: goland?  you mean the IDE?   don't understand, please reword.

mengqiy: Typo: it should be `golang`.

pkg/kinflate/util/labelsandannotations.go
```diff
@@ -33,19 +33,19 @@ type MapTransformer struct {
 var _ Transformer = &MapTransformer{}
 
 // NewDefaultingLabelsMapTransformer construct a MapTransformer with defaultLabelsPathConfigs.
-func NewDefaultingLabelsMapTransformer(m map[string]string) (Transformer, error) {
+func NewDefaultingLabelsMapTransformer(m map[string]string) (*MapTransformer, error) {
```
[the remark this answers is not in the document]
mengqiy: I was thinking about doing something like this.
```
concreteTypeTransformer, _ := NewDefaultingLabelsMapTransformer(m map[string]string)
func TestSomething() {
  // I can reference func (o *MapTransformer) addMap(in interface{}) (interface{}, error) with a concrete type.
  // concreteTypeTransformer.addMap()
}
```
But after thinking it more, I agree you are right. We should not return a concrete type. I will revert this piece of change. :)

pkg/kinflate/configmap_secret.go
mengqiy: To be more concrete, here is an example: https://play.golang.org/p/zcW07XETcsB

mengqiy: Will remove this function. We don't really need to worry about it.

Issue #227, opened 2018-01-26
Consider preferring USERPROFILE as the home path more often on Windows


dls314:
**Is this a request for help?**: No

**What keywords did you search in Kubernetes issues before filing this one?** : homedrive

---

**Is this a BUG REPORT or FEATURE REQUEST?** (choose one): bug-feature report-request

**Kubernetes version** (use `kubectl version`): 1.8.2

**Environment**:
- **Cloud provider or hardware configuration**: Docker for Windows, Minikube
- **OS** (e.g. from /etc/os-release): Windows 10
- **Kernel** (e.g. `uname -a`): n/a
- **Install tools**: n/a
- **Others**: n/a

**What happened**:
**What you expected to happen**:
**How to reproduce it** (as minimally and precisely as possible):
**Anything else we need to know**:

The `.kube` directory and `.kube\config` are being located inconsistently among some set of tooling including Docker for Windows CE 18.02.0-ce-rc1-win50 (15489) and kubectl 1.8.2.

I noticed this because `kubectl config get-clusters` returned an empty list, but I expected to see `docker-for-desktop` and `minikube` based on the contents from my `%USERPROFILE%\.kube\config` file.

I'm in a corporate environment where HOMEDRIVE HOMEPATH and HOMESHARE are in use to redirect home directories to a network drive. I only mention this because it leads to these paths differing from USERPROFILE.

In minikube, I think that the preference/fallback order leads to picking USERPROFILE as in this code:
https://github.com/kubernetes/minikube/blob/master/vendor/github.com/docker/machine/libmachine/mcnutils/utils.go#L30-L33

In kubectl, I think that the preference/fallback order leads to picking HOMEDRIVE + HOMEPATH as in this code:
https://github.com/kubernetes/kubectl/blob/master/pkg/pluginutils/plugin_client.go#L41-L44

In docker ce, I think that the preference/fallback order leads to picking USERPROFILE as in this code:
https://github.com/docker/docker-ce/blob/master/components/engine/pkg/homedir/homedir_windows.go

Those are only my best guesses about relevant code, and there are lots of other places in the wider codebases where you can find `os.Getenv("USERPROFILE")` or `os.Getenv("HOMEPATH")`, but for this narrow combination, based on behavior, I think that kubectl is the odd one out.

I think a different fallback order that prefers USERPROFILE as the other tools seem to would be nice.

dls314:
As far as I know, the relevant windows feature here is best described by this topic:
[...]

PR #228, opened 2018-01-26
Update plugin_client.go

dls314:
Prefer USERPROFILE as the home path if available.

Possible fix for
https://github.com/kubernetes/kubectl/issues/227

### DIFF
--- a/pkg/pluginutils/plugin_client.go
+++ b/pkg/pluginutils/plugin_client.go
@@ -38,9 +38,9 @@ func InitConfig() (*restclient.Config, error) {
 	// to ~/.kube/config as a last resort.
 	home := os.Getenv("HOME")
 	if runtime.GOOS == "windows" {
-		home := os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
+		home := os.Getenv("USERPROFILE")
 		if home == "" {
-			home = os.Getenv("USERPROFILE")
+			home = os.Getenv("HOMEDRIVE") + os.Getenv("HOMEPATH")
 		}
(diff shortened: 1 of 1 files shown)


dls314:
/assign @pwittrock
Suggested by @k8s-ci-robot 

/assign @jberkhahn
Last editor in plugin_client.go

adohe:
@dls314 any reason why should we change this? I am not that familiar with windows

dls314:
Consistency with minikube and docker ce, they  use the USERPROFILE value. Links to them in #227

jberkhahn:
I am also not too familiar with windows. Poking around in the links you had in your open issue, it does appear that Minikube and Docker prefer USERPROFILE. However, note that this PR changes a library used for finding and initializing your kubeconfig from within a kubectl plugin, and is not the place kubectl does this itself. I don't know which environment variable is used for that.

That said, this LGTM, but I'm not actually a member here yet.

dls314:
> However, note that this PR changes a library used for finding and initializing your kubeconfig from within a kubectl plugin, and is not the place kubectl does this itself.

That makes me think I missed something that should be included with this PR. Do you know where kubectl does this itself?

jberkhahn:
Somewhere in (kubernetes/kubernetes)[https://github.com/kubernetes/kubernetes/]. I haven't worked on the actual CLI much.

dls314:
I'd believe it might be here, anyone care to comment?
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/client-go/util/homedir/homedir.go

PR #232, opened 2018-01-30
add mock fs and file for testing

mengqiy:


mengqiy:
Forgot to run `go fmt` after fixing golint. Pushed a new commit.

PR #235, opened 2018-01-30
kinflate: Create AddConfigMap structure

apelisse:
This is empty and does nothing.

### DIFF
--- a/pkg/kinflate/commands/configmap.go
+++ b/pkg/kinflate/commands/configmap.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package commands
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+type addConfigMap struct {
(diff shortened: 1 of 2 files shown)


pkg/kinflate/commands/configmap.go
```diff
@@ -0,0 +1,77 @@
...
+package commands
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/spf13/cobra"
+)
+
+type AddConfigMap struct {
```
monopole: private?

pkg/kinflate/commands/configmap.go
```diff
@@ -0,0 +1,77 @@
...
+		RunE: func(_ *cobra.Command, args []string) error {
+			err := config.Validate()
+			if err != nil {
+				return err
+			}
+
+			if len(args) != 1 {
+				return fmt.Errorf("error: exactly one NAME is required, got %d", len(args))
+			}
+			config.Name = args[0]
```
monopole: suggest Validate accepts args in line 56 so you have something to validate.

also, please add test

pkg/kinflate/commands/configmap.go
```diff
@@ -0,0 +1,77 @@
...
+type AddConfigMap struct {
+	// Name of configMap (required)
+	Name string
+	// FileSources to derive the configMap from (optional)
+	FileSources []string
+	// LiteralSources to derive the configMap from (optional)
+	LiteralSources []string
+	// EnvFileSource to derive the configMap from (optional)
+	EnvFileSource string
+}
```
monopole: suggest reference the Generic struct in pkg/apis/manifest/v1alpha1/types.go

pkg/kinflate/commands/configmap.go
apelisse: Yeah, we'll add tests later

pkg/kinflate/commands/configmap.go
apelisse: I don't think we want to strongly couple this structure with the definition of the manifest (nor have a comment that will be obsolete in the near future).

PR #239, opened 2018-01-31
Apelisse droot seans3

apelisse:


pkg/kinflate/commands/configmap_test.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	for _, test := range tests {
+		if test.config.Validate([]string{"name"}) == nil && test.shouldFail {
+			t.Fatalf("Validation should fail if %s", test.name)
+		} else if test.config.Validate([]string{"name"}) != nil && !test.shouldFail {
+			t.Fatalf("Validation should succeed if %s", test.name)
+		}
+	}
+}
```
monopole: like this very much, succinct handling of complexity

PR #240, opened 2018-02-01
kinflate: Manifest type improvements

apelisse:
This also updates the code, the tests and the example.


pkg/kinflate/examples/simple/instances/exampleinstance/Kube-manifest.yaml
```diff
@@ -22,19 +22,15 @@ patches:
...
+- name: app-env
   env: configmap/app.env
-- type: file
-  namePrefix: app-config
+- name: app-config
   files:
   - configmap/app-init.ini
 #There could be secrets in Base, if just using a fork/rebase workflow
-secrets:
-- type: tls
```
ant31: What's the reasoning to  move out from `type: 'tls`' and use a top-level key?

ant31:
Where are the discussions made about fields and such? is there a public proposal/document to review? 
cc @monopole

apelisse:
Sorry @ant31, discussions happened off-line. I'm happy to hear your opinion though. It's unclear from your message if you think it's a bad idea, and if yes, why?

I don't think it's a good practice to have many different exclusive values within the same type. In this case it's: either use these 3 generic "source" fields (which have their own validation logic) or use just this other TLS field (that also has its validation logic). That's usually a sign that the type is trying to do too many things.

I'm a little worried about having too many top-level fields though. We could also have a top-level "secrets" key, with "tls" and "generic" being sub-keys if needed.

WDYT?

ant31:
@apelisse I think it would end-up with too many top-level key. 
I'm expecting the Manifest to become the 'declarative' representation of the current kubectl generators:
```
  clusterrole         Create a ClusterRole.
  clusterrolebinding  Create a ClusterRoleBinding for a particular ClusterRole
  configmap           Create a configmap from a local file, directory or literal
value
  deployment          Create a deployment with the specified name.
  namespace           Create a namespace with the specified name
  role                Create a role with single rule.
  rolebinding         Create a RoleBinding for a particular Role or ClusterRole
  secret              Create a secret using specified subcommand
  service             Create a service using specified subcommand.
  serviceaccount      Create a service account with the specified name
```
And that's already a bunch of top-level keys.
 
I would prefer the use of sub-keys, as you suggested:

```
secrets:
  tls: 
  - name: app-tls
    certFile: secret/tls.cert
    keyFile: secret/tls.key
 ```


note related to this issue: 
Would it be possible to track the discussions on a public document? So we can understand the history, changes and the current state of the proposal.

apelisse:
> I'm expecting the Manifest to become the 'declarative' representation of the current kubectl generators:
>```
>  clusterrole         Create a ClusterRole.
>  clusterrolebinding  Create a ClusterRoleBinding for a particular ClusterRole
>  configmap           Create a configmap from a local file, directory or literal
> value
>  deployment          Create a deployment with the specified name.
>  namespace           Create a namespace with the specified name
>  role                Create a role with single rule.
>  rolebinding         Create a RoleBinding for a particular Role or ClusterRole
>  secret              Create a secret using specified subcommand
>  service             Create a service using specified subcommand.
>  serviceaccount      Create a service account with the specified name
>```

Either of us is misunderstanding what "the Manifest" will be. Only configmap and secret are basically created from external files (non-resources) with a `--from-file` argument. IOW, I don't think we'll have more resources than we have today (or not the order of magnitude that you're pointing at).

PR #241, opened 2018-02-01
multitransformer

mengqiy:


pkg/kinflate/util/multitransformer.go
```diff
@@ -0,0 +1,39 @@
...
+
+package util
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+// MultiTransformer contains a list of transformers.
+type MultiTransformer struct {
+	Transformers []Transformer
```
monopole: Lets continue to avoid public members.

This object needs a constructor that returns an immutable object, meaning the incoming slice should be copied.  something like

```
func NewMultiTransformer(arg []Transformer) Transformer {
	r := &MultiTransformer{
		transformers: make([]Transformer, len(arg))}
	copy(r.transformers, arg)
	return r
}
```
alternatively, one could have an Append method, but no reason to have  both.

monopole: yes, Append would make the object mutable, but in a simple obvious way - more constrained than having an underlying slice that could change at any time.

PR #256, opened 2018-02-06
Add configmap

droot:
This PR adds "configmap" command to kinflate.

### DIFF
--- a/pkg/kinflate/util/manifestloader.go
+++ b/pkg/kinflate/util/manifestloader.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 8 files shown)


pkg/kinflate/util/manifestloader.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		return nil, err
+	}
+	var manifest manifest.Manifest
+	err = yaml.Unmarshal(bytes, &manifest)
+	if err != nil {
+		return nil, err
+	}
+	dir, _ := path.Split(filename)
+	adjustPathsForManifest(&manifest, []string{dir})
```
apelisse: Just a quick note: Reading and then Writing is not without side-effects because of this very line. `adjustPaths` changes the path in the manifest (which is fine if it's just an internal representation, but we actually write this one back so it's not...) based on where the command is run.

We need to move `adjustPath` somewhere else. It doesn't belong here.

droot: Good catch. I agree that need to find a better place for adjustPath..
Its not causing any side effects currently because we are assuming manifest file in current directory (adjustPaths... is a no-op).

PR #257, opened 2018-02-06
Kinflate add generic secret command.

seans3:


pkg/kinflate/commands/secret.go
```diff
@@ -63,6 +78,35 @@ func newCmdAddSecretGeneric(errOut io.Writer) *cobra.Command {
...
+	return nil
+}
+
+func getOrCreateGenericSecret(m *manifest.Manifest, name string) *manifest.GenericSecret {
+	for i, s := range m.GenericSecrets {
+		if name == s.Name {
+			return &m.GenericSecrets[i]
+		}
+	}
+	// config map not found, create new one and add it to the manifest.
```
apelisse: update comment

pkg/kinflate/commands/secret.go
```diff
@@ -63,6 +78,35 @@ func newCmdAddSecretGeneric(errOut io.Writer) *cobra.Command {
...
+	gs := &manifest.GenericSecret{Name: name}
+	m.GenericSecrets = append(m.GenericSecrets, *gs)
```
apelisse: why do you get a pointer, if the goal is to push a copy? (that probably already applied to configmap ...)

pkg/kinflate/commands/secret.go
```diff
@@ -45,12 +51,21 @@ func newCmdAddSecretGeneric(errOut io.Writer) *cobra.Command {
 				return err
 			}
 
-			if len(args) != 1 {
-				return fmt.Errorf("error: exactly one NAME is required, got %d", len(args))
+			// Load in the manifest file.
```
apelisse: Not a big fan of this kind of comments :-)

pkg/kinflate/commands/secret.go
seans3: Fixed

pkg/kinflate/commands/secret.go
seans3: Fixed

PR #270, opened 2018-02-08
manifest tree builder

mengqiy:
There will be another PR to update the transformers and switch kinflate inflate to use the tree builder.

### DIFF
--- a/pkg/kinflate/configmapandsecret/configmap_secret.go
+++ b/pkg/kinflate/configmapandsecret/configmap_secret.go
@@ -28,8 +28,8 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
 	cutil "k8s.io/kubectl/pkg/kinflate/configmapandsecret/util"
-	"k8s.io/kubectl/pkg/kinflate/gvkn"
 	"k8s.io/kubectl/pkg/kinflate/hash"
+	"k8s.io/kubectl/pkg/kinflate/types"
 )
 
 // MakeConfigmapAndGenerateName makes a configmap and returns the configmap and the name appended with a hash.
@@ -175,27 +175,27 @@ func validateTLS(cert, key string) error {
 	return nil
 }
 
-func populateMap(m map[gvkn.GroupVersionKindName]*unstructured.Unstructured, obj *unstructured.Unstructured, newName string) error {
+func populateMap(m types.KObject, obj *unstructured.Unstructured, newName string) error {
 	accessor, err := meta.Accessor(obj)
 	if err != nil {
 		return err
 	}
 	oldName := accessor.GetName()
 	gvk := obj.GetObjectKind().GroupVersionKind()
-	gvkn := gvkn.GroupVersionKindName{GVK: gvk, Name: oldName}
+	gvkn := types.GroupVersionKindName{GVK: gvk, Name: oldName}
 
 	if _, found := m[gvkn]; found {
-		return fmt.Errorf("cannot use a duplicate name %q for %s", oldName, gvk)
+		return fmt.Errorf("The <name: %q, GroupVersionKind: %v> already exists in the map", oldName, gvk)
 	}
 	accessor.SetName(newName)
 	m[gvkn] = obj
 	return nil
 }
 
-// MakeMapOfConfigMap returns a map of <GVK, oldName> -> unstructured object.
-func MakeMapOfConfigMap(manifest *manifest.Manifest) (map[gvkn.GroupVersionKindName]*unstructured.Unstructured, error) {
-	m := map[gvkn.GroupVersionKindName]*unstructured.Unstructured{}
-	for _, cm := range manifest.Configmaps {
+// MakeConfigMapsKObject returns a map of <GVK, oldName> -> unstructured object.
+func MakeConfigMapsKObject(maps []manifest.ConfigMap) (types.KObject, error) {
+	m := types.KObject{}
+	for _, cm := range maps {
 		unstructuredConfigMap, nameWithHash, err := MakeConfigmapAndGenerateName(cm)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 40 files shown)


pkg/kinflate/configmapandsecret/configmap_secret.go
```diff
@@ -172,3 +174,68 @@ func validateTLS(cert, key string) error {
 	}
 	return nil
 }
+
```
monopole: not in the PR, but i'd like to see us introduce

```
type GVKNObject map[gvkn.GroupVersionKindName]*unstructured.Unstructured
```
and start using it.  Lots of methods we could hang there.

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+	m, err := manifestPathToManifest(path)
+	if err != nil {
+		return nil, err
+	}
+	return manifestToManifestNode(m)
+}
+
+func manifestToManifestNode(m *manifest.Manifest) (*ManifestNode, error) {
```
Liujingfang1: Can you add some comment for this function? Thank you.

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+		return nil, err
+	}
+	dir, _ := path.Split(filename)
+	adjustpath.AdjustPathsForManifest(&m, []string{dir})
+	return &m, err
+}
+
+// validateManifestPath loads the manifest from the given path.
+// It returns ManifestData and an potential error.
+func validateManifestPath(mPath string) (string, error) {
```
droot: we can probably move this function to manifestloader.go

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	TLSSecrets, err := cutil.MakeMapOfTLSSecret(m)
+	err = mergemap.Merge(mdata.Secrets, TLSSecrets)
+	if err != nil {
+		return nil, err
+	}
+	return mdata, nil
+}
+
+func pathsToMap(paths []string) (map[gvkn.GroupVersionKindName]*unstructured.Unstructured, error) {
+	res := map[gvkn.GroupVersionKindName]*unstructured.Unstructured{}
```
droot: I think we have type defined for this right ? ResourceTypes ?

droot: and lot of other places also.

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	mdata.Configmaps, err = cutil.MakeMapOfConfigMap(m)
+	if err != nil {
+		return nil, err
+	}
+	mdata.Secrets, err = cutil.MakeMapOfGenericSecret(m)
+	if err != nil {
+		return nil, err
+	}
+	TLSSecrets, err := cutil.MakeMapOfTLSSecret(m)
+	err = mergemap.Merge(mdata.Secrets, TLSSecrets)
```
droot: I think here we are coupling the transformation with loading manifest data. Don't know if de-coupling is in scope of this PR, but we need to decouple the transformation from traversal.

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+		if err != nil {
+			return nil, err
+		}
+		mnode.Children = append(mnode.Children, child)
+	}
+	return mnode, nil
+}
+
+// manifestPathToManifest loads a manifest file and parse it in to the Manifest object.
+func manifestPathToManifest(filename string) (*manifest.Manifest, error) {
```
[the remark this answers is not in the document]
mengqiy: Just saw that. Will update this func.

pkg/kinflate/tree/node.go
```diff
@@ -0,0 +1,56 @@
...
+type ManifestData struct {
+	Name              string
+	NamePrefix        NamePrefixType
+	ObjectLabels      ObjectLabelsType
+	ObjectAnnotations ObjectAnnotationsType
+	Resources         ResourcesType
+	Patches           PatchesType
+	Configmaps        ConfigmapsType
+	Secrets           SecretsType
+}
```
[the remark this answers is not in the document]
mengqiy: IMO we need or will need to distinguish the resources come from subpackage and current manifest.

pkg/kinflate/tree/builder.go
mengqiy: No transformation (i.e. nameprefix, labels and annotations) has happened when building the tree.

mengqiy: What does your `transformation` mean here?

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	if f.IsDir() {
+		return fmt.Errorf("%q is NOT expected to be an dir", filename)
+	}
+	content, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+
+	_, err = kutil.Decode(content, into)
```
monopole: Does this complain about name collisions, i.e. name already in the map?

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	TLSSecrets, err := cutil.MakeMapOfTLSSecret(m)
+	err = mergemap.Merge(mdata.Secrets, TLSSecrets)
+	if err != nil {
+		return nil, err
+	}
+	return mdata, nil
+}
+
+func pathsToMap(paths []string) (map[gvkn.GroupVersionKindName]*unstructured.Unstructured, error) {
```
monopole: loadKObjectFromPaths(paths []string) (KObject, error) {...}

pkg/kinflate/tree/builder.go
droot: I have some ideas here, lets discuss.

pkg/kinflate/tree/builder.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+	return res, nil
+}
+
+func pathToMap(path string, into map[gvkn.GroupVersionKindName]*unstructured.Unstructured) error {
+	_, err := os.Stat(path)
+	if err != nil {
+		return err
+	}
+	if into == nil {
```
monopole: This should be an error.

Alternatively, this method doesn't have to take KObject as an arg.  It could just create one, and the caller could merge it into some other KObject.

pkg/kinflate/tree/builder.go
mengqiy: No. Will add a check in `func Decode()`

mengqiy:
I have dropped the commit that moves `Adjustpath`.

PR #274, opened 2018-02-10
Kinflate use manifest tree builder

mengqiy:


pkg/kinflate/transformers/overlay.go
```diff
@@ -0,0 +1,81 @@
...
+func (o *OverlayTransformer) Transform(baseResourceMap types.KObject) error {
+	// Strategic merge the resources exist in both base and overlay.
+	for gvkn, base := range baseResourceMap {
+		// Merge overlay with base resource.
+		if overlay, found := o.overlay[gvkn]; found {
+			versionedObj, err := scheme.Scheme.New(gvkn.GVK)
+			if err != nil {
+				switch {
+				case runtime.IsNotRegisteredError(err):
+					return fmt.Errorf("CRD and TPR are not supported now: %v", err)
```
droot: How do we know the resource involved is TPR/CRD ?

pkg/kinflate/transformers/overlay.go
```diff
@@ -0,0 +1,81 @@
...
+			merged, err := strategicpatch.StrategicMergeMapPatch(
+				base.UnstructuredContent(),
+				overlay.UnstructuredContent(),
+				versionedObj)
+			if err != nil {
+				return err
+			}
+			base.SetName(baseName)
+			baseResourceMap[gvkn].Object = merged
+			delete(o.overlay, gvkn)
```
droot: This means the overlay can be applied once. I think we should add a note somewhere. This can cause subtle bugs.

droot: Curious if we have multiple overlay defined for one resource, how are we doing the merge (and conflicts) ?

pkg/kinflate/tree/default.go
```diff
@@ -0,0 +1,73 @@
...
+// 4) apply annotations
+// 5) update name reference
+func DefaultTransformer(m *ManifestData) (transformers.Transformer, error) {
+	ts := []transformers.Transformer{}
+
+	ot, err := transformers.NewOverlayTransformer(types.KObject(m.Patches))
+	if err != nil {
+		return nil, err
+	}
+	if ot != nil {
```
droot: Shouldn't this be an error ?
Generally, if we didn't return err above, we should not get an empty transformer.

droot: same comment below.

pkg/kinflate/tree/node.go
```diff
@@ -20,39 +20,66 @@ import (
...
-type manifestData struct {
-	name              string
-	namePrefix        namePrefixType
-	objectLabels      objectLabelsType
-	objectAnnotations objectAnnotationsType
-	resources         resourcesType
-	patches           patchesType
-	configmaps        configmapsType
-	secrets           secretsType
+func (md *ManifestData) AllResources() error {
```
droot: Do we need to export this fn ?

pkg/kinflate/tree/node.go
```diff
@@ -20,39 +20,66 @@ import (
...
+	return types.Merge(md.Resources, md.Secrets)
+}
+
+func (md *ManifestData) Inflate() error {
+	for _, pkg := range md.Packages {
+		err := pkg.Inflate()
+		if err != nil {
+			return err
+		}
+		err = types.Merge(md.Resources, pkg.Resources)
```
droot: May be separate this merge from pkg-inflation ?

merge them at the end when all the sub-packages have been inflated ? (so we don't cause any side-effect in case of errors)

pkg/kinflate/examples/simple/instances/exampleinstance/Kube-manifest.yaml
```diff
@@ -15,8 +15,8 @@ objectLabels:
   repo: test-infra
 objectAnnotations:
   note: This is a test annotation
-resources:
-- ../../package
+packages:
```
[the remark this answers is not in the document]
mengqiy: I believe we need to migrate them.

pkg/kinflate/transformers/overlay.go
mengqiy: When it failed to find the schema.

I will improve the error message a bit.

pkg/kinflate/transformers/overlay.go
mengqiy: After discussing with @droot offline, we agreed to do this in a followup PR to keep behavior changes out of this PR.

pkg/kinflate/tree/default.go
mengqiy: Will addresses in a followup PR.

PR #279, opened 2018-02-13
error out when a patch has not matching object

mengqiy:
Address some comments from https://github.com/kubernetes/kubectl/pull/274:
- error out when a patch has not matching object
- it stop deleting the object from the overlay KObject.


pkg/kinflate/transformers/overlay.go
```diff
@@ -43,39 +43,32 @@ func NewOverlayTransformer(overlay types.KObject) (Transformer, error) {
 // Transform apply the overlay on top of the base resources.
 func (o *OverlayTransformer) Transform(baseResourceMap types.KObject) error {
 	// Strategic merge the resources exist in both base and overlay.
-	for gvkn, base := range baseResourceMap {
+	for gvkn, overlay := range o.overlay {
```
droot: +1 it should speed things up :)

droot: And this also takes care of overlay deletion problem. Nice.

PR #282, opened 2018-02-13
WIP: fix filesystem tech debt

Liujingfang1:
Replaced the os.Stat by filesystem.Stat in builder.go
Improved the error message when loading resources

### DIFF
--- a/pkg/kinflate/commands/inflate.go
+++ b/pkg/kinflate/commands/inflate.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/kubectl/pkg/kinflate/tree"
 	"k8s.io/kubectl/pkg/kinflate/types"
 	kutil "k8s.io/kubectl/pkg/kinflate/util"
+	"k8s.io/kubectl/pkg/kinflate/util/fs"
 )
 
 type inflateOptions struct {
@@ -82,7 +83,8 @@ func (o *inflateOptions) Complete(cmd *cobra.Command, args []string) error {
 // RunKinflate runs inflate command (do real work).
 func (o *inflateOptions) RunKinflate(out, errOut io.Writer) error {
 	// Build a tree of ManifestData.
-	root, err := tree.LoadManifestDataFromPath(o.manifestPath)
+	loader := tree.Loader{FS: fs.MakeRealFS(), Path: o.manifestPath}
+	root, err := loader.LoadManifestDataFromPath(o.manifestPath)
 	if err != nil {
 		return err
 	}
--- a/pkg/kinflate/tree/builder.go
+++ b/pkg/kinflate/tree/builder.go
@@ -18,7 +18,6 @@ package tree
 
 import (
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 
@@ -29,31 +28,36 @@ import (
 	"k8s.io/kubectl/pkg/kinflate/util/fs"
 )
 
+type Loader struct {
+	FS   fs.FileSystem
+	Path string
+}
+
 // LoadManifestDataFromPath takes a path to a Kube-manifest.yaml or a dir that has a Kube-manifest.yaml.
 // It returns a tree of ManifestData.
-func LoadManifestDataFromPath(path string) (*ManifestData, error) {
-	return loadManifestDataFromPath(path)
+func (l *Loader) LoadManifestDataFromPath(path string) (*ManifestData, error) {
+	return l.loadManifestDataFromPath(path)
 }
 
 // loadManifestDataFromPath make a ManifestData from path
-func loadManifestDataFromPath(path string) (*ManifestData, error) {
-	m, err := loadManifestFileFromPath(path)
+func (l *Loader) loadManifestDataFromPath(path string) (*ManifestData, error) {
+	m, err := l.loadManifestFileFromPath(path)
 	if err != nil {
 		return nil, err
 	}
-	return manifestToManifestData(m)
+	return l.manifestToManifestData(m)
 }
 
 // manifestToManifestData make a ManifestData given an Manifest object
-func manifestToManifestData(m *manifest.Manifest) (*ManifestData, error) {
-	mdata, err := loadManifestDataFromManifestFileAndResources(m)
+func (l *Loader) manifestToManifestData(m *manifest.Manifest) (*ManifestData, error) {
+	mdata, err := l.loadManifestDataFromManifestFileAndResources(m)
 	if err != nil {
 		return nil, err
 	}
 
 	pkgs := []*ManifestData{}
 	for _, pkg := range m.Packages {
-		pkgNode, err := loadManifestDataFromPath(pkg)
+		pkgNode, err := l.LoadManifestDataFromPath(pkg)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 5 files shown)


pkg/kinflate/commands/inflate.go
```diff
@@ -82,7 +83,8 @@ func (o *inflateOptions) Complete(cmd *cobra.Command, args []string) error {
 // RunKinflate runs inflate command (do real work).
 func (o *inflateOptions) RunKinflate(out, errOut io.Writer) error {
 	// Build a tree of ManifestData.
-	root, err := tree.LoadManifestDataFromPath(o.manifestPath)
+	loader := tree.Loader{FS: fs.MakeRealFS(), Path: o.manifestPath}
+	root, err := loader.LoadManifestDataFromPath(o.manifestPath)
```
apelisse: why does the loader need the path twice?

pkg/kinflate/util/manifestloader.go
```diff
@@ -33,6 +33,10 @@ type ManifestLoader struct {
 	FS fs.FileSystem
 }
 
+//This is a placeholder for the ManifestErrors type
+//TODO: Think about this struct and make is better
+type ManifestErrors []string
```
apelisse: shouldn't this be a list of errors rather than strings?

pkg/kinflate/tree/builder.go
```diff
@@ -64,30 +68,38 @@ func manifestToManifestData(m *manifest.Manifest) (*ManifestData, error) {
...
-func loadManifestFileFromPath(filename string) (*manifest.Manifest, error) {
-	loader := kutil.ManifestLoader{fs.MakeRealFS()}
-	m, err := loader.Read(filename)
+func (l *Loader) loadManifestFileFromPath(filename string) (*manifest.Manifest, error) {
+	var loader = kutil.ManifestLoader{FS: l.FS}
+	fullFilename, err := loader.MakeValidManifestPath(filename)
+	if err != nil {
+		return nil, err
+	}
+	l.Path = fullFilename
```
apelisse: I'm a little bit confused by this variable. Who is supposed to set it? who is supposed to read it? in what order?

pkg/kinflate/tree/builder.go
```diff
@@ -64,30 +68,38 @@ func manifestToManifestData(m *manifest.Manifest) (*ManifestData, error) {
...
 	mdata.NamePrefix = NamePrefixType(m.NamePrefix)
 	mdata.ObjectLabels = m.ObjectLabels
 	mdata.ObjectAnnotations = m.ObjectAnnotations
 
-	res, err := loadKObjectFromPaths(m.Resources)
+	res, err := l.LoadKObjectFromPaths(m.Resources)
 	if err != nil {
-		return nil, err
+		errorMsg := fmt.Sprintf("Resource from Manifest (%s) couldn't be loaded properly.\n%v\n"+
+			"Please check the Resource subsection in (%s).", l.Path, err, l.Path)
```
apelisse: This is using the Path in the error, but only in the error ... :confused:

pkg/kinflate/tree/builder.go
```diff
@@ -144,21 +156,21 @@ func loadKObjectFromPath(path string, into types.KObject) error {
...
-		err = loadKObjectFromFile(filepath, into)
+		err = l.LoadKObjectFromFile(filepath, into)
 		return nil
 	})
 	return e
 }
 
-func loadKObjectFromFile(filename string, into types.KObject) error {
-	f, err := os.Stat(filename)
+func (l *Loader) LoadKObjectFromFile(filename string, into types.KObject) error {
```
apelisse: is this exporting a lot more symbols? It's hard to know, by looking at this change, what is the impact on the godoc though

pkg/kinflate/util/manifestloader_test.go
```diff
@@ -62,21 +62,21 @@ func TestLoadNotExist(t *testing.T) {
 	fakeFS.Mkdir(".", 0644)
 	fakeFS.Create(badSuffix)
 	loader := kutil.ManifestLoader{FS: fakeFS}
-	_, err := loader.Read("Kube-manifest.yaml")
+	_, err := loader.MakeValidManifestPath("Kube-manifest.yaml")
 	if err == nil {
 		t.Fatalf("expect an error")
 	}
 	if !strings.Contains(err.Error(), "Run `kinflate init` first") {
```
apelisse: Not your fault, but relying on exact output string for errors makes is creating a maintenance burden.

pkg/kinflate/tree/builder.go
```diff
@@ -64,30 +68,38 @@ func manifestToManifestData(m *manifest.Manifest) (*ManifestData, error) {
...
+	fullFilename, err := loader.MakeValidManifestPath(filename)
+	if err != nil {
+		return nil, err
+	}
+	l.Path = fullFilename
+	m, err := loader.Read(fullFilename)
 	if err != nil {
 		return nil, err
 	}
+	loader.Validate(m)
```
apelisse: I know this function is a placeholder for now, but shouldn't it return an error and have the error handled?

pkg/kinflate/tree/builder_test.go
```diff
@@ -321,8 +329,9 @@ func TestLoadManifestDataFromPath(t *testing.T) {
 	parent1.Packages = []*ManifestData{child1}
 	parent2.Packages = []*ManifestData{child2}
 
+	loader := Loader{FS: fs.MakeRealFS()}
 	expected := grandparent
-	actual, err := loadManifestDataFromPath("testdata/hierarchy")
+	actual, err := loader.loadManifestDataFromPath("testdata/hierarchy")
```
apelisse: Should we test "private" methods?

pkg/kinflate/util/manifestloader.go
```diff
@@ -97,3 +97,25 @@ func (m *ManifestLoader) Write(filename string, manifest *manifest.Manifest) err
 
 	return m.fs().WriteFile(filename, bytes)
 }
+
+// Read must have already been called and we have a loaded manifest
+func (m *ManifestLoader) Validate(manifest *manifest.Manifest) ManifestErrors {
```
apelisse: why not just return `[]error`?

pkg/kinflate/commands/inflate.go
Liujingfang1: We added a `Path` member to the Loader struct with the aim that it will help us improving the error message. We don't need the second `o.manifestPath`, I will remove it.

pkg/kinflate/util/manifestloader.go
Liujingfang1: `ManifestErrors` is a struct that we are going to define and use as the data store for all the errors we can get when loading a manifest.  It should be easy to add, group and print the erros. A list of errors could serve the basic need, but we want to think about a better struct.

pkg/kinflate/tree/builder.go
Liujingfang1: There is no need to make this method public. I will change it.

pkg/kinflate/tree/builder.go
Liujingfang1: I will move this change to a different PR and update it there.

pkg/kinflate/tree/builder_test.go
Liujingfang1: I think so. It doesn't matter if the method is private or not. We need some tests to cover the code.

pkg/kinflate/util/manifestloader.go
Liujingfang1: Will move this to a different PR.

pkg/kinflate/tree/builder.go
seans3: Yes, this is confusing. We have a Loader abstraction which creates a ManifestLoader object. We intend to clean this up by integrating Loader and ManifestLoader into a coherent whole. But we don't want to do that in this initial PR. So basically, the ManifestLoader (called loader) calculated the full manifest file path. Then we store this value in the Loader object (l).

pkg/kinflate/tree/builder.go
seans3: This will be enhanced once we've fleshed out the ManifestError struct.

PR #285, opened 2018-02-14
kinflate: Generate secrets with commands

apelisse:
Remove the DataSources for secret, and generate the content of secrets
with the output of commands. This allows more dynamic encrypted/secure
secrets that are not just saved plain in the repository.

Things that are not perfect with this code:
- I'm wondering if we should just merge back TLS secret with generic secrets now. I think that TLS secrets should also be generated by commands. It would re-simplify the code and the Kube-manifest file.
- I've removed many tests, the coverage for the new code is definitely not as good,
- I'm not sure this is going to work flawlessly when there are multiple Kube-manifest files in the process. Are the commands going to be executed from where they should? ... Do we have a good way to test that use-case?


apelisse:
Related to #286. I'll re-merge tlsSecret and genericSecrets now.

apelisse:
Updated, with tlsSecret and GenericSecrets merged. Should fix #286 altogether.

pkg/kinflate/examples/simple/instances/exampleinstance/Kube-manifest.yaml
```diff
@@ -28,9 +28,11 @@ configmaps:
   files:
   - configmap/app-init.ini
 #There could be secrets in Base, if just using a fork/rebase workflow
-tlsSecrets:
+secrets:
```
monopole: secretMaker or whatever.  srlsy.

pkg/kinflate/configmapandsecret/configmap_secret.go
```diff
@@ -224,18 +170,13 @@ func MakeGenericSecretsKObject(secrets []manifest.GenericSecret) (types.KObject,
...
-			return nil, err
-		}
+func createSecretKey(wd string, command string) ([]byte, error) {
+	fi, err := os.Stat(wd)
+	if err != nil || !fi.IsDir() {
+		wd = filepath.Dir(wd)
 	}
-	return m, nil
+	cmd := exec.Command("sh", "-c", command)
+	cmd.Dir = wd
```
monopole: suppose my secret generator command is "sleep 5000"
or the command could try to run a program that isn't installed, etc.

need timeout and error handling, e.g. start a go routine the timeouts after a while, and this code selects on it or command completion.  The test could literally be a sleep longer than whatever you select as a timeout.

want to do in this PR or file an issue to add it later?

droot: CommandContext fits this use case pretty well.
https://golang.org/pkg/os/exec/#CommandContext

apelisse: I'll update now.

pkg/apis/manifest/v1alpha1/types.go
```diff
@@ -142,29 +136,24 @@ type ConfigMap struct {
 	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
 }
 
-// GenericSecret contains the metadata of how to generate a generic secret.
-type GenericSecret struct {
+// Secret contains the metadata of how to generate a generic secret.
+type Secret struct {
```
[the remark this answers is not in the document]
apelisse: I don't know if I agree. This still has the list of secret names in there. It used to be secret before, even though it was built from literal/file/etc. After-all, the struct looks like this from a user-perspective:
```
secret:
    one-secret:
        key1: command1
        key2: command2
    other-secret:
        key1: command1
        key2: command2     
```

Or do you like this more:
```
secretGenerator:
    one-secret:
        key1: command1
        key2: command2
    other-secret:
        key1: command1
        key2: command2     
```
?

pkg/kinflate/examples/simple/instances/exampleinstance/Kube-manifest.yaml
apelisse: Yeah, that answers my question =))

pkg/kinflate/configmapandsecret/configmap_secret.go
```diff
@@ -224,18 +172,15 @@ func MakeGenericSecretsKObject(secrets []manifest.GenericSecret) (types.KObject,
...
-		if err != nil {
-			return nil, err
-		}
+func createSecretKey(wd string, command string) ([]byte, error) {
+	fi, err := os.Stat(wd)
+	if err != nil || !fi.IsDir() {
+		wd = filepath.Dir(wd)
 	}
-	return m, nil
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
monopole: Betcha $1 the 5 sec becomes 30 before Bastille Day

pkg/kinflate/configmapandsecret/configmap_secret_test.go
```diff
@@ -252,84 +179,36 @@ func TestConstructConfigMap(t *testing.T) {
...
-				DataSources: manifest.DataSources{
-					LiteralSources: []string{"a=x", "b=y"},
-				},
-			},
-			expected: makeLiteralSecret("literalSecret"),
+func TestFailConstructSecret(t *testing.T) {
+	secret := manifest.SecretGenerator{
+		Name: "secret",
+		Commands: map[string]string{
+			"FAILURE": "false", // This will fail.
```
monopole: Nice comment.  It's not obvious that there's a command called false that always fails :)

Does the coreutils on a Mac have this?  I don't care.

pkg/kinflate/configmapandsecret/configmap_secret_test.go
```diff
@@ -252,84 +179,36 @@ func TestConstructConfigMap(t *testing.T) {
...
-				CertFile: "../examples/simple/instances/exampleinstance/secret/tls.cert",
-				KeyFile:  "../examples/simple/instances/exampleinstance/secret/tls.key",
-			},
-			expected: makeTLSSecret("tlsSecret"),
+func TestConstructSecret(t *testing.T) {
+	secret := manifest.SecretGenerator{
+		Name: "secret",
+		Commands: map[string]string{
+			"DB_USERNAME": "printf admin",
+			"DB_PASSWORD": "printf somepw",
```
monopole: Sweet test

pkg/kinflate/configmapandsecret/configmap_secret_test.go
apelisse: Yeah, that's kind of mandatory, otherwise you can't do `if false; then ...` which is obviously super useful =)
FYI, there is also `true` that always succeeds :-)

PR #292, opened 2018-02-15
add kinflate diff command

mengqiy:
- support noop option for kinflate inflate
- add a test for noop option
- port some diff util and update vendor to make it builds.
- add kinflate diff command (ONLY diff the current layer(Node), not the whole tree)

fixes: #295

### DIFF
--- a/pkg/kinflate/commands/inflate.go
+++ b/pkg/kinflate/commands/inflate.go
@@ -31,7 +31,7 @@ import (
 
 type inflateOptions struct {
 	manifestPath string
-	namespace    string
+	mode         tree.ModeType
 }
 
 // newCmdInflate creates a new inflate command.
@@ -56,7 +56,7 @@ func newCmdInflate(out, errOut io.Writer) *cobra.Command {
 				fmt.Fprintf(errOut, "error: %v\n", err)
 				os.Exit(1)
 			}
-			err = o.RunKinflate(out, errOut)
+			err = o.RunInflate(out, errOut)
 			if err != nil {
 				fmt.Fprintf(errOut, "error: %v\n", err)
 				os.Exit(1)
@@ -66,7 +66,6 @@ func newCmdInflate(out, errOut io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVarP(&o.manifestPath, "filename", "f", "", "Pass in a Kube-manifest.yaml file or a directory that contains the file.")
 	cmd.MarkFlagRequired("filename")
-	cmd.Flags().StringVarP(&o.namespace, "namespace", "o", "yaml", "Output mode. Support json or yaml.")
 	return cmd
 }
 
@@ -77,11 +76,12 @@ func (o *inflateOptions) Validate(cmd *cobra.Command, args []string) error {
 
 // Complete completes inflate command.
 func (o *inflateOptions) Complete(cmd *cobra.Command, args []string) error {
+	o.mode = tree.ModeNormal
 	return nil
 }
 
 // RunKinflate runs inflate command (do real work).
-func (o *inflateOptions) RunKinflate(out, errOut io.Writer) error {
+func (o *inflateOptions) RunInflate(out, errOut io.Writer) error {
 	// Build a tree of ManifestData.
 	loader := tree.Loader{FS: fs.MakeRealFS(), InitialPath: o.manifestPath}
 	root, err := loader.LoadManifestDataFromPath()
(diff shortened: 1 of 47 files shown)


pkg/kinflate/commands/inflate.go
```diff
@@ -87,8 +87,13 @@ func (o *inflateOptions) RunKinflate(out, errOut io.Writer) error {
...
 	}
 
+	tGetter := tree.DefaultTransformer
+	if o.noop {
+		tGetter = tree.NoOpTransformer
+	}
+
 	// Do the transformation for the tree.
-	err = root.Inflate()
+	err = root.Inflate(tGetter)
```
droot: hmm. Can inflate take an bool arg noOp ? and hide this details from the caller ?

pkg/kinflate/transformers/mergekobject.go
```diff
@@ -0,0 +1,41 @@
...
+	if len(kobj) == 0 {
+		return NewNoOpTransformer(), nil
+	}
+	return &MergeKObjectTransformer{kobj: kobj}, nil
+}
+
+// Transform merges 2 KObjects
+func (o *MergeKObjectTransformer) Transform(m types.KObject) error {
+	return types.Merge(m, o.kobj)
+}
```
droot: Not very clear about the motivation....why would this be a transformer ?

pkg/kinflate/tree/default.go
```diff
@@ -30,6 +30,17 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	ts = append(ts, mt1)
+	mt2, err := transformers.NewMergeKObjectTransformer(types.KObject(m.Secrets))
+	if err != nil {
+		return nil, err
+	}
+	ts = append(ts, mt2)
+
```
droot: hmm.... want to think more if ...if this is the right way to approach it.

pkg/kinflate/commands/inflate.go
```diff
@@ -65,7 +65,7 @@ func newCmdInflate(out, errOut io.Writer) *cobra.Command {
 
 	cmd.Flags().StringVarP(&o.manifestPath, "filename", "f", "", "Pass in a Kube-manifest.yaml file or a directory that contains the file.")
 	cmd.MarkFlagRequired("filename")
-	cmd.Flags().StringVarP(&o.namespace, "namespace", "o", "yaml", "Output mode. Support json or yaml.")
+	cmd.Flags().BoolVarP(&o.noop, "noop", "", false, "kinflate will do no transformation if true.")
```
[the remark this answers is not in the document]
mengqiy: I thought it may be helpful for the users.
But if you don't agree, I can just remove it.

pkg/kinflate/commands/inflate.go
mengqiy: That works for now.

pkg/kinflate/transformers/mergekobject.go
mengqiy: Every single manipulation on the KObject is now a transformer.
No need to do preprocessing:
```
Merge(Resources, ConfigMaps)
Merge(Resources, Secrets)
Transform(Resources)
```

pkg/kinflate/tree/node.go
```diff
@@ -80,10 +81,19 @@ func (md *ManifestData) Inflate() error {
...
+		t = transformers.NewNoOpTransformer()
+	} else {
+		err = md.allResources()
+		if err != nil {
+			return err
+		}
+		t, err = DefaultTransformer(md)
+		if err != nil {
+			return err
+		}
```
monopole: please use something like 
```
func makeTransformer(noop bool, md whatever) (Transformer, err) {
  if noop {
   return NewNoop...
 }
  return the other thing
}
```

pkg/kinflate/commands/inflate.go
```diff
@@ -30,7 +30,7 @@ import (
 
 type inflateOptions struct {
 	manifestPath string
-	namespace    string
+	noop         bool
```
monopole: this is fine for now.  We'll have an enum eventually, e.g.
```
  mode ModeType
```
where
```
  const (
	modeUnknown ModeType = iota
	ModeNoop
        ModeLint
        ModeNormal
        ???
)
```

mengqiy:
Updated. PTAL.

pkg/kinflate/commands/testdata/testcase-noop/mysql/service.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: Service
+metadata:
+  name: mysql
+  labels:
+    app: mysql
+spec:
+  ports:
+    - port: 3306
+  selector:
+    app: mysql
```
apelisse: Missing new line :-)

mengqiy:
Updated. PTAL.

PR #296, opened 2018-02-15
kinflate: removed non-existing resources from manifest file

droot:
`kinflate init` generates a manifest file which contains
non-existent resources in `resources` section. This change
keeps those resources in commented form so that 'kinflate inflate'
doesnt barf at the user.

### DIFF
--- a/pkg/kinflate/commands/commands.go
+++ b/pkg/kinflate/commands/commands.go
@@ -31,9 +31,9 @@ func NewDefaultCommand() *cobra.Command {
 
 	c := &cobra.Command{
 		Use:   "kinflate",
-		Short: "kinflate is a kubernetes cluster configuration utility",
+		Short: "kinflate manages Kubernetes application configuration.",
 		Long: `
-kinflate is a kubernetes cluster configuration utility.
+kinflate manages Kubernetes application configuration.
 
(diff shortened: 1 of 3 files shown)


pkg/kinflate/commands/commands.go
```diff
@@ -31,9 +31,9 @@ func NewDefaultCommand() *cobra.Command {
 
 	c := &cobra.Command{
 		Use:   "kinflate",
-		Short: "kinflate is a kubernetes cluster configuration utility",
+		Short: "kinflate is a commandline tool to manage Kubernetes application configurations.",
```
monopole: Since its obviously a commandline tool, suggest simply
```
kinflate manages Kubernetes application configuration.
```

pkg/kinflate/commands/commands.go
```diff
@@ -31,9 +31,9 @@ func NewDefaultCommand() *cobra.Command {
...
 		Long: `
-kinflate is a kubernetes cluster configuration utility.
+kinflate is a commandline tool to manage Kubernetes application configurations.
```
monopole: ditto

droot: Done.

PR #301, opened 2018-02-16
Resource Library

ellenkorbes:
Package resource implements tools for the discovery and filtering of resources in an API server.

Tests to be implemented in a subsequent PR.


pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+	→ deployments
+	→→ &{Resource:{Name:deployments SingularName: Namespaced:true Group:v1beta1 Version: Kind:Deployment Verbs:[create delete deletecollection get list patch update watch] ShortNames:[deploy] Categories:[all]} ApiGroupVersion:extensions/v1beta1 Schema:0xc420ed7400 SubResources:[]}
+	→→ &{Resource:{Name:deployments SingularName: Namespaced:true Group:v1beta1 Version: Kind:Deployment Verbs:[create delete deletecollection get list patch update watch] ShortNames:[deploy] Categories:[all]} ApiGroupVersion:apps/v1beta1 Schema:0xc4202e9400 SubResources:[]}
+	...
+
+Filtering can also be applied by implementing the Filter interface.
+
+The following (admittedly ludicrous) example adds a filter to the example above that excludes all resources that do not start with the letter "n":
+
+	type letterN struct {
```
apelisse: Yeah, there is no reason not to have this as a compiled example.

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+*/
+
+/*
+Package resource implements tools for the discovery and filtering of resources in an API server.
+
+The starting point should be the creation of a new Parser object, which can then use the Resources() method to discover resources in the API server.
+
+For example:
+
+	p := resource.NewParser(resources, discovery, apiGroup, apiVersion)
```
apelisse: Same, I think this could be compiled. We can't test the output, but we can make sure it compiles.

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+
+Filtering can also be applied by implementing the Filter interface.
+
+The following (admittedly ludicrous) example adds a filter to the example above that excludes all resources that do not start with the letter "n":
+
+	type letterN struct {
+	}
+
+	func (*letterN) Resource(r *resource.Resource) bool {
+		return string(r.Resource.Name[0]) == "n"
```
apelisse: It'd be weird to have an empty resource name, but that could technically crash ;-)

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+	...
+
+Filtering can also be applied by implementing the Filter interface.
+
+The following (admittedly ludicrous) example adds a filter to the example above that excludes all resources that do not start with the letter "n":
+
+	type letterN struct {
+	}
+
+	func (*letterN) Resource(r *resource.Resource) bool {
```
apelisse: Doesn't need to be pointer.

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+		return true
+	}
+
+	func example() {
+		p := resource.NewParser(resources, discovery, apiGroup, apiVersion)
+		r, err := p.Resources()
+		if err != nil {
+			panic(err)
+		}
+		r = r.Filter(&letterN{})
```
apelisse: Doesn't have to be address (if you remove the two pointers above)

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+
+	func (*letterN) Resource(r *resource.Resource) bool {
+		return string(r.Resource.Name[0]) == "n"
+	}
+
+	func (*letterN) SubResource(*resource.SubResource) bool {
+		return true
+	}
+
+	func example() {
```
apelisse: That's a very cool example ;-)

pkg/framework/resource/doc.go
```diff
@@ -0,0 +1,78 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+/*
+Package resource implements tools for the discovery and filtering of resources in an API server.
+
+The starting point should be the creation of a new Parser object, which can then use the Resources() method to discover resources in the API server.
```
apelisse: It's true for all your comments throughout the pull-request: try to limit the number of characters to 70-ish per line.

pkg/framework/resource/filter.go
```diff
@@ -0,0 +1,92 @@
...
+package resource
+
+// Filter is an interface whose methods are used to arbitrarily filter resources and subresources in a Resources object. Filtering criteria can be anything at all, so long as the "accepted" resources return true, and the filtered out resources return false.
+type Filter interface {
+	Resource(*Resource) bool
+	SubResource(*SubResource) bool
+}
+
+// NewAndFilter that takes as argument a slice of filters, which then work as a logical AND. This filter will only return resources that pass through every filter in the argument slice. It returns a new Filter value.
+func NewAndFilter(filters ...Filter) Filter {
```
apelisse: Can you replace the content of this function (and apply the same logic for or) with:
```
return &andFilter{Filters: filters}
```

pkg/framework/resource/filter.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// NewOrFilter that takes as argument a slice of filters, which then work as a logical OR. This filter will return resources that pass through any one or more filters in the argument slice. It returns a new Filter value.
+func NewOrFilter(filters ...Filter) Filter {
+	new := orFilter{}
+	new.Filters = append(new.Filters, filters...)
+	return &new
+}
+
+type emptyFilter struct {
```
apelisse: This is definied privately, but not used in this file. I don't know if it's used anywhere else in the package, but I still find this a little dubious

EDIT: It's used nowhere.

pkg/framework/resource/parse.go
```diff
@@ -0,0 +1,211 @@
...
+		return nil, err
+	}
+	resources, byGVR := p.indexResources(gvs)
+	err = p.attachSubResources(gvs, resources, byGVR)
+	return resources, err
+}
+
+// subResource returns a resource name, subresource name pair, and true if the resource is a subresource.
+// It returns a resource name, an empty string, and false if the resource is not a subresource.
+func (*parser) subResource(resource *v1.APIResource) (string, string, bool) {
```
apelisse: since the meaning of these return parameters can't be inferred by their type, you should maybe name them.

pkg/framework/resource/parse.go
```diff
@@ -0,0 +1,211 @@
...
+	parts := strings.Split(resource.Name, "/")
+	if len(parts) > 1 {
+		return parts[0], parts[1], true
+	}
+	return parts[0], "", false
+}
+
+// resource returns the resource name and true if it is a resource (not a subresource).
+// It returns the resource name and false it if it is a subresource.
+func (p *parser) resource(resource *v1.APIResource) (string, bool) {
```
apelisse: naming return parameters?

pkg/framework/resource/parse.go
```diff
@@ -0,0 +1,211 @@
...
+	if len(resource.Group) == 0 {
+		resource.Group = group
+	}
+	if len(resource.Version) == 0 {
+		resource.Group = version
+	}
+}
+
+// isGroupVersionMatch returns false if either group or version doesn't match with the API.
+func (p *parser) isGroupVersionMatch(group, version string) bool {
```
apelisse: I'm wondering why we need that? If we could get rid of that, that'd simplify the creation of the parser. @pwittrock

pkg/framework/resource/resource.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	openapi "k8s.io/kube-openapi/pkg/util/proto"
+)
+
+// Resource is an API Resource.
+type Resource struct {
+	Resource        v1.APIResource
```
apelisse: I think we'd rather make these private by default, and change them to public as needed.

pkg/framework/resource/resource.go
```diff
@@ -0,0 +1,59 @@
...
+	ApiGroupVersion schema.GroupVersion
+	Schema          openapi.Schema
+	SubResources    []*SubResource
+}
+
+// SubResource is an API subresource.
+type SubResource struct {
+	Resource        v1.APIResource
```
apelisse: public -> private?

pkg/framework/resource/parse.go
```diff
@@ -0,0 +1,211 @@
...
+func (p *parser) getOpenAPI(group, version, kind string) (proto.Schema, bool) {
+	schema := p.resources.LookupResource(schema.GroupVersionKind{group, version, kind})
+	if schema == nil {
+		return nil, false
+	}
+	return schema, true
+}
+
+// indexResources indexes into maps the resources from the API resource list. It returns a map indexed by resource name, and a map indexed by GroupVersionResource objects.
+func (p *parser) indexResources(gvs []*v1.APIResourceList) (map[string][]*Resource, map[schema.GroupVersionResource]*Resource) {
```
apelisse: Should the type be "Resources" here (first returned type)?

apelisse:
Yeah, you should run `gofmt` too.

pkg/framework/resource/parse.go
```diff
@@ -0,0 +1,211 @@
...
+	if len(p.apiGroup) > 0 && p.apiGroup != group {
+		return false
+	}
+	if len(p.apiVersion) > 0 && p.apiVersion != version {
+		return false
+	}
+	return true
+}
+
+// getOpenAPI retrieves a schema object from the API based on a GroupVersionResource triplet.
```
apelisse: I'm not sure why this returns the Schema and bool. It's safe and sufficient to check if Schema is nil.

pkg/framework/resource/filter.go
ellenkorbes: Should we delete it or make it public?

pkg/framework/resource/resource.go
ellenkorbes: You sure? If I were to do this it'd break my filter examples (they depend on r.Resource.Name).

I tried it anyway and the IDE gave me:

> Rename failed: /home/ellen/go/src/k8s.io/kubectl/pkg/framework/resource/resource.go:26:6: renaming this type "Resource" to "resource" would make it unexported /home/ellen/go/src/k8s.io/kubectl/pkg/framework/resource/poc/main.go:38:38: breaking references from packages such as "k8s.io/kubectl/pkg/framework/resource/poc" [...]

I did it by hand then and indeed it broke everything.

> undefined: "k8s.io/kubectl/pkg/framework/resource".Resource

There's probably a way to do the filtering differently, but do we really have to?

pkg/framework/resource/doc.go
ellenkorbes: Since the examples won't pass testing, isn't that possibly gonna create a problem for us later on?

pkg/framework/resource/resource.go
apelisse: Sorry if I was unclear. By "these", I meant "all these fields". If this one in particular is needed, let's keep it public, what about the others?

pkg/framework/resource/doc.go
apelisse: if you don't have the `// Output:` comment in your example, they won't run, they won't fail. They will be compiled to make sure that they are up-to-date though (which is still valuable to us).

apelisse:
This is great. Please fix the test so that we can merge it!

Thanks!

PR #310, opened 2018-02-21
update yaml import pkg

mengqiy:
make all the yaml pkgs we use consist.

### DIFF
--- a/pkg/kinflate/commands/inflate_test.go
+++ b/pkg/kinflate/commands/inflate_test.go
@@ -25,7 +25,7 @@ import (
 	"strings"
 	"testing"
 
-	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


pkg/kinflate/commands/inflate_test.go
```diff
@@ -25,7 +25,7 @@ import (
 	"strings"
 	"testing"
 
-	"gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
```
droot: ha ha :)

mengqiy:
> Do you know why this one is better than the other?

`github.com/ghodss/yaml` is the wrapper of `gopkg.in/yaml.v2` to better handle some problems.
This is the only place using `gopkg.in/yaml.v2`, we should make it consist with others.

PR #313, opened 2018-02-22
added configmap resource implementation

droot:
This is a very early idea of implementing resource for configmap objects just to get a handle on the shape of the solution.

### DIFF
--- a/pkg/kinflate/resource/configmap.go
+++ b/pkg/kinflate/resource/configmap.go
@@ -1,0 +1,67 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resource
+
+import (
+	corev1 "k8s.io/api/core/v1"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	cutil "k8s.io/kubectl/pkg/kinflate/configmapandsecret/util"
+)
+
+// (Note): pass in loader which has rootPath context and knows how to load
+// files given a relative path.
+// NewFromConfigMap returns a Resource given a configmap metadata from manifest
+// file.
+func NewFromConfigMap(cm manifest.ConfigMap) (*Resource, error) {
+	corev1CM, err := makeConfigMap(cm)
+	if err != nil {
+		return nil, err
+	}
+
+	data, err := objectToUnstructured(corev1CM)
+	if err != nil {
+		return nil, err
+	}
+	return &Resource{Data: data}, nil
+}
+
+func makeConfigMap(cm manifest.ConfigMap) (*corev1.ConfigMap, error) {
+	corev1cm := &corev1.ConfigMap{}
+	corev1cm.APIVersion = "v1"
+	corev1cm.Kind = "ConfigMap"
+	corev1cm.Name = cm.Name
+	corev1cm.Data = map[string]string{}
+
+	// TODO: move the configmap helpers functions in this file/package
+	if cm.EnvSource != "" {
+		if err := cutil.HandleConfigMapFromEnvFileSource(corev1cm, cm.EnvSource); err != nil {
+			return nil, err
+		}
+	}
+	if cm.FileSources != nil {
+		if err := cutil.HandleConfigMapFromFileSources(corev1cm, cm.FileSources); err != nil {
+			return nil, err
+		}
+	}
+	if cm.LiteralSources != nil {
+		if err := cutil.HandleConfigMapFromLiteralSources(corev1cm, cm.LiteralSources); err != nil {
+			return nil, err
+		}
+	}
+
+	return corev1cm, nil
+}
--- a/pkg/kinflate/resource/resource.go
+++ b/pkg/kinflate/resource/resource.go
@@ -16,10 +16,41 @@ limitations under the License.
 
 package resource
 
-import "k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+import (
+	"encoding/json"
 
-// Examples: deployment, service, configmap, secret, etc.
-type Resource interface {
-	Name() string
-	Data() unstructured.Unstructured
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/kinflate/types"
+)
+
+// Resource represents a Kubernetes Resource Object for ex. Deployment, Server
+// ConfigMap etc.
+type Resource struct {
+	Data *unstructured.Unstructured
+}
+
+// GVKN returns Group/Version/Kind/Name for the resource.
+func (r *Resource) GVKN() types.GroupVersionKindName {
+	var emptyZVKN types.GroupVersionKindName
+	if r.Data == nil {
+		return emptyZVKN
+	}
+	accessor, err := meta.Accessor(r.Data)
+	if err != nil {
+		return emptyZVKN
+	}
+	gvk := r.Data.GetObjectKind().GroupVersionKind()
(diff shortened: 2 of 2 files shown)


pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+	cutil "k8s.io/kubectl/pkg/kinflate/configmapandsecret/util"
+	"k8s.io/kubectl/pkg/kinflate/hash"
+)
+
+// configMap implements Resource interface for ConfigMap type.
+type configMap struct {
+	cm manifest.ConfigMap
+	// add loader here
+
+	name string
```
apelisse: I know it's wip, but this is not needed :-)

pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+	data *unstructured.Unstructured
```
apelisse: Do we need the cache? isn't it early optimization?

pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,112 @@
...
+func makeConfigMap(cm manifest.ConfigMap) (*corev1.ConfigMap, error) {
+	corev1cm := &corev1.ConfigMap{
+		APIVersion: "v1",
+		Kind:       "ConfigMap",
+		Name:       cm.Name,
+		Data:       map[string]string{},
+	}
+
+	if cm.EnvSource != "" {
+		if err := cutil.HandleConfigMapFromEnvFileSource(corev1cm, cm.EnvSource); err != nil {
```
apelisse: do we want to move that cutil code closer in this resource package?

pkg/kinflate/resource/configmap.go
droot: Taking it out :)

pkg/kinflate/resource/configmap.go
droot: Yes, we do.

pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,80 @@
...
+	if err != nil {
+		return nil, err
+	}
+	data, err := objectToUnstructured(corev1CM)
+	if err != nil {
+		return nil, err
+	}
+	return &Resource{
+		Data:       data,
+		NameSuffix: h,
```
apelisse: Do we need this? Can't we have the name already include the suffix?

pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,80 @@
...
+	corev1cm.Data = map[string]string{}
+
+	// TODO: move the configmap helpers functions in this file/package
+	if cm.EnvSource != "" {
+		if err := cutil.HandleConfigMapFromEnvFileSource(corev1cm, cm.EnvSource); err != nil {
+			return nil, err
+		}
+	}
+	if cm.FileSources != nil {
+		if err := cutil.HandleConfigMapFromFileSources(corev1cm, cm.FileSources); err != nil {
```
apelisse: Note for later: we might want to pass `FileSources...` to `HandleConfigMapFromFileSources` (after changing its signature of course) and then we don't have to check for nil. (and same below)

pkg/kinflate/resource/resource.go
```diff
@@ -16,10 +16,57 @@ limitations under the License.
...
+
+// Resource represents a Kubernetes Resource Object for ex. Deployment, Server
+// ConfigMap etc.
+type Resource struct {
+	Data *unstructured.Unstructured
+
+	// This is primarily to unify the processing workflow for configmap with
+	// other resources. For configmap, we use hash of the payload as suffix for
+	// the name. This will be empty for other resources.
+	NameSuffix string
```
apelisse: That looks like a special case in a generic struct :-/

pkg/kinflate/resource/resource.go
```diff
@@ -16,10 +16,57 @@ limitations under the License.
...
+}
+
+func (r *Resource) Name() string {
```
apelisse: I don't think we need the name if we have the GVKN

pkg/kinflate/resource/configmap.go
droot: Discussed with @mengqiy , we need the old-name of the resources because we need to update references for old-name after transforming.

So may be we can call OldName which is applicable to all the resources anyways and update the name in the Data.

pkg/kinflate/resource/configmap.go
droot: Agreed. I was planning to do it while integrating with loader.

pkg/kinflate/resource/resource.go
droot: Agreed, look at my previous comment.

pkg/kinflate/resource/resource.go
droot: agreed. will kill it.

droot:
@apelisse updated the PR PTAL.

pkg/kinflate/resource/resource.go
```diff
@@ -16,10 +16,46 @@ limitations under the License.
...
+
+// Resource represents a Kubernetes Resource Object for ex. Deployment, Server
+// ConfigMap etc.
+type Resource struct {
+	Data *unstructured.Unstructured
+
+	// OrigName is the name of this resource at the time of of loading it from a
+	// source. Actual name may be differ because we add suffix for configmaps
+	// or name-prefix.
+	OrigName string
```
apelisse: I really don't want to block, but this still seem quite wrong. That is an un-pumped special case in a general structure.

Can we append the hash to the config map as a transformation? Then we give a chance for whoever created this resource to save the name. e.g.

```
r, _ := NewFromConfigMap(cm) // Create the resource from manifest.ConfigMap
oldName := r.GVKN().Name // Save its name
r := AppendHashName(r) // Append the hash
newName := r.GVKN().Name // New name
// s/oldName/newName/ where needed.
```

WDYT?

Liujingfang1: Considering the usage of this `OrigName` for different resources(resource, configmap, secret, patch),  it can be used to change the name from the oldName to newName. Currently only configmap and secret need to have newName, which is with hash suffix. If we don't have this OrigName, we can still append that suffix.

droot: @apelisse @Liujingfang1  I like the idea of handling suffix as a transformer. What that means is: we can combine all the name transformations like name-prefix/hash(only for configmap) in to one transformer. I am liking it even more now. @mengqiy do you see any problem with that ?

droot: Removed origName for now. We will try to handle it in the name transformation. If that doesn't work, then we will add it back.

pkg/kinflate/resource/resource.go
```diff
@@ -16,10 +16,41 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/kubectl/pkg/kinflate/types"
+)
+
+// Resource represents a Kubernetes Resource Object for ex. Deployment, Server
+// ConfigMap etc.
+type Resource struct {
+	Data *unstructured.Unstructured
```
apelisse: Do we need a pointer?

pkg/kinflate/resource/configmap.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	data, err := objectToUnstructured(corev1CM)
+	if err != nil {
+		return nil, err
+	}
+	return &Resource{Data: data}, nil
+}
+
+func makeConfigMap(cm manifest.ConfigMap) (*corev1.ConfigMap, error) {
```
mengqiy: This func is copied from somewhere else in kinflate.
Purely copy or with any change?

pkg/kinflate/resource/resource.go
mengqiy: I believe we do. Unstructured is a struct instead of an interface.

pkg/kinflate/resource/configmap.go
droot: This is copied and has been changed now. The plan is to eventually get rid of the original code.

pkg/kinflate/resource/resource.go
apelisse: That doesn't necessarily means you should have a pointer though, does it? :-)
I'm curious if there's a rule for that ...

droot: One thing to note here is having it as struct doesn't help much because it has map[string]interface{} as a field inside where all the data resides and that is NOT copied when you pass the struct around. So technically you are NOT working on a copy.

PR #314, opened 2018-02-22
Add appresource implementation

Liujingfang1:


pkg/kinflate/resource/appresource.go
```diff
@@ -0,0 +1,76 @@
...
+package resource
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	kutil "k8s.io/kubectl/pkg/kinflate/util"
+	"k8s.io/kubectl/pkg/kinflate/util/fs"
+)
+
+type appResource struct {
```
apelisse: Yeah, I think this will be covered by #313

pkg/kinflate/resource/appresource.go
```diff
@@ -0,0 +1,76 @@
...
+func (ap *appResource) Data() unstructured.Unstructured {
+	return *ap.data
+}
+
+type appResourceList struct {
+	path string
+	fsys fs.FileSystem
+}
+
+func NewAppResourceList(path string) *appResourceList {
```
apelisse: Yeah, eventually will take the loader and the path

pkg/kinflate/resource/appresource.go
```diff
@@ -0,0 +1,76 @@
...
+	return &appResourceList{path: path, fsys: fs.MakeRealFS()}
```
apelisse: Lazy-loading, ok

pkg/kinflate/resource/appresource.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// We will use loader.load to replace filesystem.Stat and filesystem.ReadFile
+func (ap *appResourceList) Resources() ([]appResource, error) {
```
apelisse: Maybe we should just make this a function tbh

`NewResourcesFromAppResource(loader, path) ([]*Resources, error)`? (`Resources` being what Sunil has defined in #313).

pkg/kinflate/util/util.go
```diff
@@ -50,6 +48,17 @@ func Decode(in []byte, into types.KObject) (types.KObject, error) {
 	if err != io.EOF {
 		return nil, err
 	}
+	return objs, nil
+}
+
+// DecodeToKObject decodes a list of objects in byte array format.
+// Decoded object will be inserted in `into` if it's not nil. Otherwise, it will
+// construct a new map and return it.
+func DecodeToKObject(in []byte, into types.KObject) (types.KObject, error) {
```
apelisse: Just curiosity, do we plan on keeping `KObject`?

pkg/kinflate/resource/appresource.go
Liujingfang1: Agree. We don't need to redefine the struct.

pkg/kinflate/resource/appresource.go
Liujingfang1: I agree. We will use the resource defined in #313 and remove the appResource defined here.

pkg/kinflate/util/util.go
Liujingfang1: I don't think we need to. The type resource defined in #313 has a method GVKN(). With that, we can get the same info stored in KObject.

pkg/kinflate/resource/appresource.go
Liujingfang1: Added loader to the function parameters. Now the function is `ResourcesFromPath(path string, loader Loader)`

pkg/kinflate/resource/appresource.go
```diff
@@ -0,0 +1,40 @@
...
+*/
+
+package resource
+
+import (
+	kutil "k8s.io/kubectl/pkg/kinflate/util"
+	"k8s.io/kubectl/pkg/loader"
+)
+
+func ResourcesFromPath(path string, loader loader.Loader) ([]*Resource, error) {
```
apelisse: Name is inconsistent with https://github.com/kubernetes/kubectl/blob/master/pkg/kinflate/resource/configmap.go#L29 :-/

pkg/kinflate/resource/appresource_test.go
```diff
@@ -0,0 +1,63 @@
...
+metadata:
+  name: dply1
+---
+apiVersion: v1
+kind: Deployment
+metadata:
+  name: dply2
+`)
+
+type fakeLoader struct {
```
apelisse: Yeah, we should have a `k8s.io/kubectl/pkg/loader/loadertest`. Let's keep this for now.

pkg/kinflate/resource/appresource.go
Liujingfang1: Yes, this is because that the function here has a different behavior. It returns a slice of Resource. While in configmap.go, it returns one Resource.

apelisse: Yeah, this doesn't show up nicely in the godoc either because of that. We might want to have "Resources" on top of resources (which would help checking for conflicts on top of that). And then we could have consistent creation of resources (even for configmap we could return a list of 1 resource, but then it'd be consistent). I don't know. Too early to waste time on that :-)

PR #315, opened 2018-02-22
Refactor Loader interface

seans3:
* The Loader interface adds New(root) -> Loader method to clone new Loaders
* The factory becomes an internal implementation of loaderImpl
* Removes fakefileloader, since we can just create a loader with a fake filesystem.
TODO: Implement better filepath calculation


pkg/loader/fileloader_test.go
```diff
@@ -28,11 +28,11 @@ func TestFileLoaderHappyPath(t *testing.T) {
 	location := "foo"
 	content := []byte("bar")
 	fakefs.WriteFile(location, content)
-	l, err := NewFileLoader(fakefs, location)
```
apelisse: We might want to only test the external interface, and drop these tests.

pkg/loader/loader.go
```diff
@@ -16,6 +16,87 @@ limitations under the License.
 
 package loader
 
+import (
+	"fmt"
+
+	"k8s.io/kubectl/pkg/kinflate/util/fs"
+)
+
+// Loader abstracts how bytes are read for manifest, resource, patch, or other
```
apelisse: Surprisingly, the documentation mostly talks about `kinflate` specific concepts, but none of the variable/methods/class name are referring to `kinflate`.

apelisse:
Somebody changed the tide configuration? This should have been merged already ...

pkg/loader/fileloader_test.go
mengqiy: Can't agree on this.
Ideally we should have high TC on everything.

mengqiy:
I didn't change that.
@apelisse Where can I find the tide config for our repo?

pkg/loader/fileloader_test.go
apelisse: I think we should only test the implementation details if we can't test them through the public interface. But, yeah, nvm

apelisse:
I'm looking at it, but it's completely different from what we had before. They refactored everything I think :-)

apelisse:
https://github.com/kubernetes/test-infra/blob/master/prow/config.yaml#L79-L80

OK. This PR doesn't have the CL label. I don't know if the label is missing, or if we should remove this constraint.

pkg/loader/fileloader_test.go
apelisse: I'll rephrase. We hope to have a stable external interface, but we're probably going to change the implementation many times. A stronger set of external test would allow us to improve the implementation more quickly, because we wouldn't have to change any tests.

it means that:
- If the tests are broken, it's because the implementation is broken, not because it's different.
- the tests are very focused on the exact behavior provided by the interface. If we have many users of this interface across the project, they need to know precisely how it works.

These conditions means that we have more confidence when changing the code, and also that we don't have to maintain a bunch of tests that keeps breaking.

Nuance to that of course: some things are hard to test because you need to be able to mock some specific knobs in your implementation. I'm not opposed to that, but they should be the exception rather than the rules. I don't think the tests we have here are the exception.

mengqiy:
@apelisse Why the bot doesn't automate the CLA process for us?

apelisse:
Maybe it missed a hook and didn't apply the cla label. I don't know. I can't generalize out of this one event :-)

PR #317, opened 2018-02-24
Refactor Loader interface

seans3:
* Adds Root() method to Loader interface
* Exposes SchemeLoader interface
* Renames RootLoader() to Init()
* Dynamically adds SchemeLoaders to root loader in Init()
* Removes private fileLoader tests


pkg/loader/loader.go
```diff
@@ -16,87 +16,82 @@ limitations under the License.
...
 
-// RootLoader initializes the first Loader, with the initial root location.
-func RootLoader(root string, fs fs.FileSystem) Loader {
-	// TODO: Validate the root
-	return &loaderImpl{root: root, fs: fs}
+const emptyRoot = ""
+
+// Init initializes the first loader with the supported schemes.
+// Example schemes: fileLoader, httpLoader, gitLoader.
+func Init(schemes []SchemeLoader) Loader {
```
apelisse: Should this be `New` or `NewLoader`?

pkg/loader/fileloader.go
```diff
@@ -17,32 +17,75 @@ limitations under the License.
...
+	fullFilePath, err := l.FullLocation(root, location)
+	if err != nil {
+		return "", err
+	}
+	return filepath.Dir(fullFilePath), nil
 }
 
-// Join the root path with the location path.
-func (l *fileLoader) fullLocation(root string, location string) string {
+// If location is a full file path, then ignore root. If location is relative, then
```
apelisse: This documentation piece is not printed, because it's documenting a private struct. It's too bad because the exposed documentation is not very clear.

pkg/loader/loader.go
```diff
@@ -16,87 +16,82 @@ limitations under the License.
...
+}
+
+// Interface for different types of loaders (e.g. fileLoader, httpLoader, etc.)
+type SchemeLoader interface {
+	// Does this location correspond to this scheme.
+	IsScheme(root string, location string) bool
+	// Calulcate a new root.
+	Root(root string, location string) (string, error)
+	// Combines the root and path into a full location string.
+	FullLocation(root string, path string) (string, error)
```
apelisse: > This documentation piece is not printed, because it's documenting a private struct. It's too bad because the exposed documentation is not very clear.

I really don't understand what the difference is between `Root` and `FullLocation` by reading these two comments.

pkg/loader/loader.go
```diff
@@ -16,87 +16,82 @@ limitations under the License.
...
+	// Load bytes at scheme-specific location or an error.
```
apelisse: I'm assuming this is a filepath, not a dir?

pkg/loader/loader.go
```diff
@@ -16,87 +16,82 @@ limitations under the License.
...
-	fs   fs.FileSystem
-	// http client for URL loading
-	// git client for Git loading
+	root    string
+	schemes []SchemeLoader
+}
+
+// Interface for different types of loaders (e.g. fileLoader, httpLoader, etc.)
+type SchemeLoader interface {
+	// Does this location correspond to this scheme.
```
apelisse: > correspond

Maybe find a better verb?

pkg/loader/loader.go
seans3: I called it "Init" because it's the first loader, and we already have a "New" to create a Loader for a subdirectory.

apelisse:
Looks good. Can you please make sure you rebase on top of master (and hence remove the merge commit)?

Thanks!

seans3:
We will address minor naming and documentation issues later, since some are waiting on this.

PR #319, opened 2018-02-26
Add secret resource

Liujingfang1:
Create Resource from one `SecretGenerator` object.
It reads the command, executes the command and use the output as the secret data. 
When executing the command, it need a path `p`, which is the root directory containing Kube-manifest.YAML file. With #317, we can call this function by `NewFromSecretGenerator(s, loader.Root())`

### DIFF
--- a/pkg/kinflate/resource/secret.go
+++ b/pkg/kinflate/resource/secret.go
@@ -1,0 +1,71 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resource
+
+import (
+	"context"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+)
+
+// NewFromSecretGenerator takes a SecretGenerator and executes its command in directory p
+// then writes the output to a Resource object and returns a pointer to it.
+func NewFromSecretGenerator(s manifest.SecretGenerator, p string) (*Resource, error) {
+	corev1secret := &corev1.Secret{}
+	corev1secret.APIVersion = "v1"
+	corev1secret.Kind = "Secret"
+	corev1secret.Name = s.Name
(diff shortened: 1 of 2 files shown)


apelisse:
Is this WIP?

Liujingfang1:
Yes. This change only works well when the loader(though not showing up here) is a fileloader. If we are using a different loader,  executing a command from `loader.Root()` doesn't make much sense. I'd like to hear your thoughts on this part. Do we want to deal with `fileloader` only?

apelisse:
> Yes. This change only works well when the loader(though not showing up here) is a fileloader. If we are using a different loader, executing a command from loader.Root() doesn't make much sense. I'd like to hear your thoughts on this part. Do we want to deal with fileloader only?

As discussed. Let's try to execute from wherever and fail. This is far from perfect, but might work if the command doesn't depend on any remote file.

pkg/kinflate/resource/secret.go
```diff
@@ -0,0 +1,69 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+)
+
+func NewFromSecretGenerator(s manifest.SecretGenerator, p string) (*Resource, error) {
```
apelisse: Maybe add a comment. Looking at the signature, I don't know what is `p`, do you have a better name?

pkg/kinflate/resource/secret_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"encoding/base64"
+	"reflect"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+)
+
+func makeUnstructuredSecret(name string) *unstructured.Unstructured {
```
apelisse: you might want to move the content of this directly in the test to be honest ... (so that everything lives in one place).

pkg/kinflate/resource/secret.go
apelisse: It look me a while to figure out that `p` is actually the working directory :D

Liujingfang1:
I double checked the code. This is the exact current behavior.

Liujingfang1:
Add comment to function `NewFromSecretGenerator`
Update test to address Antoine's comment

pkg/kinflate/resource/secret.go
```diff
@@ -0,0 +1,71 @@
...
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+)
+
+// NewFromSecretGenerator takes a SecretGenerator and executes its command in directory p
+// then writes the output to a Resource object and returns a pointer to it.
```
apelisse: I still think that a better name would help :-)

PR #320, opened 2018-02-26
Change function name from ResourcesFromPath to NewFromPath

Liujingfang1:
change function name to be consistent with `NewFromConfigmap` and `NewFromSecretGenerator`


apelisse:
Thank you!

PR #322, opened 2018-02-26
implement Resources func for applicationImpl

mengqiy:


pkg/kinflate/resource/appresource.go
```diff
@@ -39,3 +39,15 @@ func NewFromPath(path string, loader loader.Loader) ([]*Resource, error) {
 	}
 	return res, nil
 }
+
+func NewFromPathList(pathList []string, loader loader.Loader) ([]*Resource, error) {
```
apelisse: Should we export both this one and the other one?

apelisse: Or should we move this one somewhere else?

pkg/kinflate/resource/secret_test.go
```diff
@@ -26,21 +26,19 @@ import (
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
 )
 
-func TestNewFromSecretGenerator(t *testing.T) {
-	secret := manifest.SecretGenerator{
+func makeSecretGenerator() manifest.SecretGenerator {
```
apelisse: Please no, the goal was to keep the data where it's tested (remove logic from tests).

pkg/kinflate/resource/appresource.go
mengqiy: I guess expose only one is fine.

pkg/kinflate/app/application.go
```diff
@@ -39,8 +45,67 @@ func New(loader loader.Loader) (Application, error) {
...
+	var m v1alpha1.Manifest
+	err = yaml.Unmarshal(manifestBytes, &m)
+	if err != nil {
+		return nil, err
+	}
+
+	allResources := []*resource.Resource{}
+	aggErr := &interror.ManifestErrors{}
+	for _, pkgPath := range m.Packages {
+		l, err := a.loader.New(pkgPath)
```
apelisse: why `l`?

pkg/kinflate/resource/appresource.go
apelisse: If we keep one, then maybe the signature should be:
`func NewFromPaths(loader loader.Loader, paths ...string) ([]*Resource, error)`

pkg/kinflate/app/application.go
```diff
@@ -39,8 +45,67 @@ func New(loader loader.Loader) (Application, error) {
...
+	manifestBytes, err := a.loader.Load(constants.KubeManifestFileName)
+	if err != nil {
+		return nil, err
+	}
+
+	var m v1alpha1.Manifest
+	err = yaml.Unmarshal(manifestBytes, &m)
+	if err != nil {
+		return nil, err
+	}
```
droot: May be combine the above code in a method on application called loadManifest() ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +45,67 @@ func New(loader loader.Loader) (Application, error) {
...
+
+	allResources := []*resource.Resource{}
+	aggErr := &interror.ManifestErrors{}
```
droot: May be call aggErr --> errs

pkg/kinflate/app/application.go
droot: pkgLoader or subPkgLoader or subLoader ?

pkg/kinflate/resource/configmap.go
```diff
@@ -130,3 +130,15 @@ func addKV(m map[string]string, kv kvPair) error {
...
+	allResources := []*Resource{}
+	for _, cm := range cmList {
+		res, err := NewFromConfigMap(cm, loader)
+		if err != nil {
+			return nil, err
+		}
+		allResources = append(allResources, res)
+	}
+	return allResources, nil
+}
```
droot: If we end up calling NewFromPaths.... for resources then we should change this to
`func NewFromConfigMaps(loader..., cms...)`
Move loader as the first arg in all the methods.

mengqiy:
@droot I just realized that we may need to keep the original names which is useful for applying overlay. :/

pkg/kinflate/resource/configmap.go
apelisse: Only one loader though :-)

pkg/kinflate/resource/util.go
```diff
@@ -0,0 +1,43 @@
...
+
+package resource
+
+import (
+	"fmt"
+
+	"k8s.io/kubectl/pkg/kinflate/types"
+)
+
+func ResourceSliceToResourceCollection(l []*Resource) (types.KObject, error) {
```
apelisse: Should we move this closer to the application? So we wouldn't have to make it public

pkg/kinflate/commands/inflate.go
```diff
@@ -99,11 +102,20 @@ func (o *inflateOptions) runInflate(fs fs.FileSystem) (types.KObject, error) {
...
+	if err != nil {
+		return err
+	}
+
+	allResource, err := application.Resources()
+	if err != nil {
+		return err
+	}
+
+	kobj, err := resource.ResourceSliceToResourceCollection(allResource)
```
Liujingfang1: We can remove this function call with the change in #324

mengqiy:
> Can you clarify that? Do you mean the name of a layer below the surface?

I mean the GVKN in a patch uses the original name. This is a problem when the patch is be applied to a resource that comes from the base layer and that has been transformed.
An example is wordpress-mysql bundle.

apelisse:
> An example is wordpress-mysql bundle.

Can you be even more specific? Give me an actual example. 

Nothing in `wordpress-mysql` changes the actual name of resources. But if they did (by maybe having a prefix), then I think whatever uses it, it should use the prefixed name in their patch (since they are patching the level just below, not a deeper layer).

#### I'll try to come-up with an example:

- Let's say we have a `mysql` application with a deployment name `mysql`.
- Let's say we have `wordpress-mysql` that includes the `mysql` application, and uses a `wm-` prefix. The name of the deployment created by that app will be `wm-mysql`, right?
- Let's say now we have a `wm-prod-instance` application, that includes the `wordpress-mysql` application. My prod instance needs to patch the deployment in the `wordpress-mysql` app (which actually comes from `mysql`). What name should the patch use for the deployment? My opinion is that it should be `wm-mysql`, since that's what is exposed by the layer directly under you. Are you suggesting it should be `mysql` since that's the original resource it's trying to patch?

Am I understanding the problem?

pkg/kinflate/resource/configmap.go
droot: yes, only one loader :)

mengqiy:
Still WIP. Not ready for thorough review.
Waiting for a fix for file loader from @seans3.
Will add some tests tomorrow.

pkg/kinflate/resource/secret_test.go
mengqiy: What logic do you want to remove here?

mengqiy: Discussed offline.
Will update it.

mengqiy:
PTAL

pkg/kinflate/app/application.go
```diff
@@ -17,18 +17,26 @@ limitations under the License.
...
-	"k8s.io/kubectl/pkg/kinflate/resource"
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
+	"k8s.io/kubectl/pkg/kinflate/types"
 	"k8s.io/kubectl/pkg/loader"
 )
 
 type Application interface {
-	Resources() []resource.Resource
+	// Resources returnsn the transformed resources.
```
droot: s/returnsn/returns

pkg/kinflate/app/application.go
```diff
@@ -17,18 +17,26 @@ limitations under the License.
...
+	// It contains resources from 1) untransformed resources from current manifest 2) transformed resources from sub packages
+	RawResources() (types.ResourceCollection, error)
 }
 
+var _ Application = &applicationImpl{}
+
 // Private implementation of the Application interface
 type applicationImpl struct {
-	manifest v1alpha1.Manifest
+	// manifest is used for getting transformer.
```
droot: I don't think this comment is useful. may be omit it completely. manifest is everything for an app  :)

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
 }
 
 // Resources computes and returns the resources from the manifest.
-func (a *applicationImpl) Resources() []resource.Resource {
+func (a *applicationImpl) Resources() (types.ResourceCollection, error) {
```
droot: We can improve the comment to say:
Resources computes and returns the resources for the app ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
 	// This is where all the fun happens.
 	// From the manifest create the resources.
-	return nil
+	m, err := newManifest(a.loader)
```
droot: Remove the above two comments.

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	a.manifest = *m
+
+	errs := &interror.ManifestErrors{}
+	allResourcesCollection, err := rescourceCollectionFromManifest(a.loader, m)
```
droot: we should avoid repeating the type "Collection" in our variable names (who knows we can change the type tomorrow :)).
May be call it, allResources or resources

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+		if err != nil {
+			errs.Append(err)
+			continue
+		}
+		subapp, err := New(subloader)
+		if err != nil {
+			errs.Append(err)
+			continue
+		}
+		subResourcesCollection, err := subapp.Resources()
```
droot: subResourcesCollection --> subAppResources ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+		if err != nil {
+			errs.Append(err)
+			continue
+		}
+		types.Merge(allResourcesCollection, subResourcesCollection)
```
droot: Not a big deal but I I am thinking if ResourceCollection should expose method to add another resourceCollection ? because this merge is not resulting into a new resourceCollection instead modifying the first collection ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+	}
+
+	patchesRC, err := resourceCollectionFromPaths(a.loader, m.Patches)
```
droot: patchesRC --> patches ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+	if err != nil {
+		errs.Append(err)
+		return nil, errs
+	}
+
```
droot: At this point, we should check if we have errors from generating all the resources and return an error ? In that case, we should just append error in the previous error check.

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+	}
+	a.manifest = *m
+
+	errs := &interror.ManifestErrors{}
+	allResourcesCollection, err := rescourceCollectionFromManifest(a.loader, m)
+	if err != nil {
+		return allResourcesCollection, err
+	}
+
+	for _, pkgPath := range m.Packages {
```
droot: May be we can have an internal method called "subAppResources()" to get all the subAppResources ?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+		errs.Append(err)
+		return nil, errs
+	}
+
+	t, err := a.getTransformer(patchesRC)
+	if err != nil {
+		errs.Append(err)
+		return nil, errs
+	}
+	err = t.Transform(types.KObject(allResourcesCollection))
```
droot: I am assuming this type-cast will go away once we get rid of types.KObject.

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,102 @@ func New(loader loader.Loader) (Application, error) {
...
+			continue
+		}
+		// Gather all transformed resources from subpackages.
+		subResourcesCollection, err := subapp.Resources()
+		if err != nil {
+			errs.Append(err)
+			continue
+		}
+		types.Merge(allResourcesCollection, subResourcesCollection)
+	}
```
droot: having a "subAppResources()" method will help to remove this code duplication.

pkg/kinflate/app/transform.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	ts = append(ts, at)
+
+	nrt, err := transformers.NewDefaultingNameReferenceTransformer()
+	if err != nil {
+		return nil, err
+	}
+	ts = append(ts, nrt)
+	return transformers.NewMultiTransformer(ts), nil
+}
```
droot: I don't think this deserves a file of its own :)

pkg/kinflate/app/application.go
mengqiy: I agree. But it should not be in the scope of this PR. Will do it in a followup PR.

pkg/kinflate/app/application.go
mengqiy: IMO we should return at this point if there are any error.
Continuing to do transformation doesn't make much sense.

pkg/kinflate/app/application.go
mengqiy: True.

mengqiy:
Comments addressed. PTAL

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,129 @@ func New(loader loader.Loader) (Application, error) {
...
+	if len(errs.Get()) > 0 {
+		return nil, errs
+	}
+
+	err = types.Merge(allResources, subAppResources)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
mengqiy: No.
`Resources()` loads resources, cms and secrets by invoking `rescourceCollectionFromManifest `, while `RawResources()`  only loads resources by calling `resourceCollectionFromPaths(a.loader, m.Resources)`

pkg/kinflate/app/application_test.go
```diff
@@ -0,0 +1,230 @@
...
+		t.Fatalf("Unexpected error %v", err)
+	}
+
+	if !reflect.DeepEqual(actual, expected) {
+		err = compareMap(actual, expected)
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func compareMap(m1, m2 types.ResourceCollection) error {
```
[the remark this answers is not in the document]
mengqiy: This is can generate more useful error message which can show you which obj if different.

pkg/kinflate/app/manifest.go
```diff
@@ -0,0 +1,104 @@
...
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	"k8s.io/kubectl/pkg/kinflate/constants"
+	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
+	"k8s.io/kubectl/pkg/kinflate/resource"
+	"k8s.io/kubectl/pkg/kinflate/types"
+	"k8s.io/kubectl/pkg/loader"
+)
+
+func newManifest(loader loader.Loader) (*manifest.Manifest, error) {
+	manifestBytes, err := loader.Load(constants.KubeManifestFileName)
```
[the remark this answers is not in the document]
mengqiy: No.
Loader will ensure path is a dir by truncating the file name.

pkg/kinflate/app/application.go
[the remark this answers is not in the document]
apelisse: It's a sign that the code composes poorly though :-/

More importantly though: raw shouldn't include the generated secrets/configmaps?

pkg/kinflate/app/manifest.go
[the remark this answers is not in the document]
apelisse: I think the interface of the loader has changed, and will not truncate filenames

pkg/kinflate/app/manifest.go
```diff
@@ -0,0 +1,104 @@
...
+		allResources = append(allResources, secrets...)
+	}
+	if len(errs.Get()) > 0 {
+		return nil, errs
+	}
+	return allResources, nil
+}
+
+// rescourceCollectionFromManifest returns a ResourceCollection given a manifest object
+func rescourceCollectionFromManifest(l loader.Loader, m *manifest.Manifest) (types.ResourceCollection, error) {
```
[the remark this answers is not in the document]
mengqiy: Good catch.

pkg/kinflate/app/application_test.go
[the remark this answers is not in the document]
apelisse: Sounds good. I'm suspecting we might have a surprising behavior if you don't use the same function to know if they are the same/find the errors. Can you replace with `if err := compareMap(...); err != nil {` instead of comparing using `DeepEqual`?

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,129 @@ func New(loader loader.Loader) (Application, error) {
...
-func (a *applicationImpl) Resources() []resource.Resource {
-	// This is where all the fun happens.
-	// From the manifest create the resources.
-	return nil
+func (a *applicationImpl) Resources() (types.ResourceCollection, error) {
+	m, err := newManifest(a.loader)
+	if err != nil {
+		return nil, err
+	}
+	a.manifest = *m
```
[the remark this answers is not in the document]
Liujingfang1: Good point. We can create the manifest when we create the application.

pkg/kinflate/app/application.go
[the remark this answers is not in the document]
apelisse: And I guess the question is: Can I have patches on top of my generated secrets (within the same "App")?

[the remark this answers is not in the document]
mengqiy: > Can I have patches on top of my generated secrets (within the same "App")?

Yes, users may patch the metadata of their secrets. IMO we should not disallow it. But our tutorial should say you should not patch your secrets' data field.

pkg/kinflate/app/manifest.go
```diff
@@ -0,0 +1,104 @@
...
+	allResources, err := resourcesFromManifest(l, m)
+	if err != nil {
+		return nil, err
+	}
+	return resourceCollectionFromResources(allResources)
+}
+
+// rescourceCollectionFromManifest returns a ResourceCollection given a path slice in a manifest
+func resourceCollectionFromPaths(l loader.Loader, paths []string) (types.ResourceCollection, error) {
```
apelisse: These two functions have very limited added value (beside coupling two things together :-/)

pkg/kinflate/app/application.go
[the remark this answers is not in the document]
apelisse: So should RAW print the created configmaps/secrets? (maybe you're right and it shouldn't)

pkg/kinflate/app/manifest.go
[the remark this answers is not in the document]
mengqiy: > I think the interface of the loader has changed, and will not truncate filenames

I didn't see any new change to the loader. git didn't fetch any new things.

I think the loader will detect the scheme (fs, http, git) based on the giver path or URL.
What we want to support:
- path/to/manifest/ (Supported, `Root` is path/to/manifest/)
- path/to/manifest/Kube-manifest.yaml (Supported, `Root` is path/to/manifest/)
- http://example.com/mysql/Kube-manifest.yaml (Not supported yet)

pkg/kinflate/app/application.go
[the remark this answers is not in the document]
mengqiy: >So should RAW print the created configmaps/secrets?

No ATM. `RawResources()` only affects `diff's behavior. 
What RAW should print depends on what do we want to show the users when running `kinflate diff`.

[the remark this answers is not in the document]
apelisse: Yeah, I think that's right. If we want generated configs and secrets to be printed as addition in the diff then we shouldn't provide them in RawResources. I still believe that Resources shold call RawResources though.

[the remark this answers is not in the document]
mengqiy: > I still believe that Resources should call RawResources though.

That can happen without change any existing behavior. :)

pkg/kinflate/app/application.go
```diff
@@ -39,8 +47,129 @@ func New(loader loader.Loader) (Application, error) {
 }
 
 // Resources computes and returns the resources from the manifest.
-func (a *applicationImpl) Resources() []resource.Resource {
-	// This is where all the fun happens.
-	// From the manifest create the resources.
-	return nil
+func (a *applicationImpl) Resources() (types.ResourceCollection, error) {
```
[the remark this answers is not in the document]
mengqiy: Done

pkg/kinflate/app/application_test.go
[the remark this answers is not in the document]
mengqiy: Done

PR #324, opened 2018-02-27
Modify Encode function

Liujingfang1:


pkg/kinflate/util/util.go
```diff
@@ -124,6 +117,21 @@ func Encode(in types.KObject) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// EncodeFromKObject encodes the map `in` and output the encoded objects separated by `---`.
+func EncodeFromKObject(in types.KObject) ([]byte, error) {
```
droot: Should we call it EncodeKObject(....) ?

We can also implement `Encode` method on types.KObject type, will be better I guess and eventually it will move to `ResourceCollection` ?

pkg/kinflate/util/util.go
```diff
@@ -124,6 +117,21 @@ func Encode(in types.KObject) ([]byte, error) {
...
+	gvknList := []types.GroupVersionKindName{}
+	for gvkn := range in {
+		gvknList = append(gvknList, gvkn)
+	}
+	sort.Sort(types.ByGVKN(gvknList))
```
droot: This is doing more than encoding. If it is called only from one place, I will move it to that package instead of having it in utils here.

pkg/kinflate/util/util.go
Liujingfang1: We can call it `Encode` and move it to package of `ResourceCollection`. We'll leave the `Encode(objs []*unstructured.Unstructured) ([]byte, error)` in utils here.

apelisse:
Should we get this merge?

Liujingfang1:
We don't want this merged now. It can wait until #331, #322 merged or be included in the code cleaning part.

PR #328, opened 2018-02-27
Simplified Loader.New() to always take a directory

seans3:
Allow the FakeLoader to be initialized with a directory; not just a full file path.

### DIFF
--- a/pkg/loader/loadertest/fake_loader.go
+++ b/pkg/loader/loadertest/fake_loader.go
@@ -18,7 +18,6 @@ package loadertest
 
 import (
 	"os"
-	"strings"
 
 	"k8s.io/kubectl/pkg/kinflate/util/fs"
 	"k8s.io/kubectl/pkg/loader"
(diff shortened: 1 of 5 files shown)


pkg/loader/loadertest/fake_loader.go
```diff
@@ -30,11 +31,15 @@ type FakeLoader struct {
...
-// must be an absolute path.
+// a fake file system that the Loader reads from. "initialFilePath" parameter
+// must be an absolute path, and it can be a directory if it has a
+// trailing slash: "/home/seans/project/" (dir) -- "/home/seans/project" (file)
 func NewFakeLoader(initialFilePath string) FakeLoader {
 	// Create fake filesystem and inject it into initial Loader.
 	fakefs := fs.MakeFakeFS()
+	if strings.HasSuffix(initialFilePath, "/") {
+		fakefs.Mkdir(initialFilePath, 0x777)
+	}
```
droot: I think fakeFS probably doesn't even care about the dirs. You can store a file at a path for which directory doesn't even exist in fakeFS.?

apelisse: I'm starting to wonder if we should have the fake fs create a `/tmp/something` and use the actual fs. The fake is going to have many limitations that we can't program easily.

PR #331, opened 2018-03-01
update resource package

mengqiy:
This PR will be useful for implementing Resources func for applicationImpl in https://github.com/kubernetes/kubectl/pull/322.

### DIFF
--- a/pkg/kinflate/resource/appresource.go
+++ b/pkg/kinflate/resource/appresource.go
@@ -21,8 +21,7 @@ import (
 	"k8s.io/kubectl/pkg/loader"
 )
 
-//  NewFromPath returns a Resource list given a resource path from manifest file.
-func NewFromPath(path string, loader loader.Loader) ([]*Resource, error) {
+func resourcesFromPath(loader loader.Loader, path string) ([]*Resource, error) {
 	content, err := loader.Load(path)
 	if err != nil {
 		return nil, err
@@ -38,4 +37,17 @@ func NewFromPath(path string, loader loader.Loader) ([]*Resource, error) {
 		res = append(res, &Resource{Data: obj})
 	}
 	return res, nil
+}
+
+//  NewFromPaths returns a slice of Resources given a resource path slice from manifest file.
+func NewFromPaths(loader loader.Loader, paths []string) ([]*Resource, error) {
+	allResources := []*Resource{}
+	for _, path := range paths {
+		res, err := resourcesFromPath(loader, path)
+		if err != nil {
+			return nil, err
+		}
+		allResources = append(allResources, res...)
+	}
+	return allResources, nil
 }
--- a/pkg/kinflate/resource/appresource_test.go
+++ b/pkg/kinflate/resource/appresource_test.go
@@ -21,7 +21,7 @@ func makeUnconstructed(name string) *unstructured.Unstructured {
 	}
 }
 
-func TestNewFromPath(t *testing.T) {
+func TestNewFromPaths(t *testing.T) {
 
 	resourceStr := `apiVersion: v1
 kind: Deployment
@@ -43,7 +43,7 @@ metadata:
 		{Data: makeUnconstructed("dply2")},
 	}
 
-	resources, _ := NewFromPath("/home/seans/project/deployment.yaml", l)
+	resources, _ := NewFromPaths(l, []string{"/home/seans/project/deployment.yaml"})
 	if len(resources) != 2 {
 		t.Fatalf("%#v should contain 2 appResource, but got %d", resources, len(resources))
 	}
--- a/pkg/kinflate/resource/configmap.go
+++ b/pkg/kinflate/resource/configmap.go
@@ -27,9 +27,8 @@ import (
 	"k8s.io/kubectl/pkg/loader"
 )
 
-// NewFromConfigMap returns a Resource given a configmap metadata from manifest file.
-func NewFromConfigMap(cm manifest.ConfigMap, l loader.Loader) (*Resource, error) {
-	corev1CM, err := makeConfigMap(cm, l)
+func newFromConfigMap(l loader.Loader, cm manifest.ConfigMap) (*Resource, error) {
+	corev1CM, err := makeConfigMap(l, cm)
 	if err != nil {
 		return nil, err
 	}
@@ -41,7 +40,7 @@ func NewFromConfigMap(cm manifest.ConfigMap, l loader.Loader) (*Resource, error)
 	return &Resource{Data: data}, nil
 }
 
-func makeConfigMap(cm manifest.ConfigMap, l loader.Loader) (*corev1.ConfigMap, error) {
+func makeConfigMap(l loader.Loader, cm manifest.ConfigMap) (*corev1.ConfigMap, error) {
 	var envPairs, literalPairs, filePairs []kvPair
 	var err error
 
@@ -129,4 +128,17 @@ func addKV(m map[string]string, kv kvPair) error {
 	}
 	m[kv.key] = kv.value
 	return nil
+}
+
+// NewFromConfigMaps returns a Resource slice given a configmap metadata slice from manifest file.
+func NewFromConfigMaps(loader loader.Loader, cmList []manifest.ConfigMap) ([]*Resource, error) {
+	allResources := []*Resource{}
+	for _, cm := range cmList {
+		res, err := newFromConfigMap(loader, cm)
+		if err != nil {
+			return nil, err
+		}
+		allResources = append(allResources, res)
+	}
+	return allResources, nil
 }
--- a/pkg/kinflate/resource/secret_test.go
+++ b/pkg/kinflate/resource/secret_test.go
@@ -26,37 +26,41 @@ import (
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
 )
 
-func TestNewFromSecretGenerator(t *testing.T) {
-	secret := manifest.SecretGenerator{
-		Name: "secret",
-		Commands: map[string]string{
-			"DB_USERNAME": "printf admin",
-			"DB_PASSWORD": "printf somepw",
+func TestNewFromSecretGenerators(t *testing.T) {
+	secrets := []manifest.SecretGenerator{
+		{
+			Name: "secret",
+			Commands: map[string]string{
+				"DB_USERNAME": "printf admin",
+				"DB_PASSWORD": "printf somepw",
+			},
+			Type: "Opaque",
 		},
-		Type: "Opaque",
 	}
-	re, err := NewFromSecretGenerator(secret, ".")
+	re, err := NewFromSecretGenerators(".", secrets)
 
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	expected := &unstructured.Unstructured{
-		Object: map[string]interface{}{
-			"apiVersion": "v1",
-			"kind":       "Secret",
-			"metadata": map[string]interface{}{
-				"name":              "secret",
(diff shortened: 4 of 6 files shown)


pkg/kinflate/resource/appresource.go
```diff
@@ -39,3 +38,16 @@ func NewFromPath(path string, loader loader.Loader) ([]*Resource, error) {
 	}
 	return res, nil
 }
+
+//  NewFromPaths returns a Resource slice given a resource path slice from manifest file.
+func NewFromPaths(loader loader.Loader, pathList []string) ([]*Resource, error) {
```
apelisse: Why do you use `[]string`, rather than `...string`?

pkg/kinflate/resource/appresource_test.go
```diff
@@ -43,7 +43,7 @@ metadata:
 		{Data: makeUnconstructed("dply2")},
 	}
 
-	resources, _ := NewFromPath("/home/seans/project/deployment.yaml", l)
+	resources, _ := NewFromPaths(l, []string{"/home/seans/project/deployment.yaml"})
```
apelisse: This would look pretty again

pkg/kinflate/resource/configmap.go
```diff
@@ -27,8 +27,7 @@ import (
 	"k8s.io/kubectl/pkg/loader"
 )
 
-// NewFromConfigMap returns a Resource given a configmap metadata from manifest file.
-func NewFromConfigMap(cm manifest.ConfigMap, l loader.Loader) (*Resource, error) {
+func newFromConfigMap(cm manifest.ConfigMap, l loader.Loader) (*Resource, error) {
```
apelisse: You forgot to invert the order of `cm` and `l` here (it's not consistent with the public method).

pkg/kinflate/resource/configmap.go
```diff
@@ -130,3 +129,16 @@ func addKV(m map[string]string, kv kvPair) error {
 	m[kv.key] = kv.value
 	return nil
 }
+
+// NewFromConfigMaps returns a Resource slice given a configmap metadata slice from manifest file.
+func NewFromConfigMaps(loader loader.Loader, cmList []manifest.ConfigMap) ([]*Resource, error) {
```
apelisse: `...` ?

pkg/kinflate/resource/manifestresource.go
```diff
@@ -0,0 +1,53 @@
...
+package resource
+
+import (
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
+	"k8s.io/kubectl/pkg/loader"
+)
+
+// NewFromManifest returns a Resource slice given a manifest object.
+func NewFromManifest(l loader.Loader, m *manifest.Manifest) ([]*Resource, error) {
```
apelisse: I'm wondering if this should be a method of the Application abstraction? 1 method giving you the cooked resources, 1 giving you the raw.

i.e. I'm not sure people should use this function directly, should they? Or maybe you can have this here, and then have `func (a *applicationImpl) Raw() ([]*Resource, error) { return NewFromManifest(a.loader, a.manifest) }`

pkg/kinflate/resource/resourcecollection.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	"k8s.io/kubectl/pkg/kinflate/types"
+	"k8s.io/kubectl/pkg/loader"
+)
+
+type ResourceCollection types.KObject
```
apelisse: I don't understand why we have both `ResourceCollection` and `types.KObject`.
Also I don't really see the value of these two `New` function. They just tie things together that don't necessarly have to be tied (iow, RC <-> manifest, RC <-> patch) but the only added value is converting a list of resources into a ResourceCollection (the real value is in the private method).

What about this:
```
func NewKobjectFromResources(l ...*Resources) (types.KObject, error)
```
as an interface?

pkg/kinflate/resource/appresource.go
Liujingfang1: When we call this function from the top level, we always have `[]string`. So we only export this function and make the argument as a slice of string.

apelisse: Yeah it doesn't really matter. It's still easy to call though from the top level `NewFromPath(a.loader, a.manifest.resources...)`. But yes, it doesn't matter.

pkg/kinflate/resource/manifestresource.go
Liujingfang1: This idea is in alignment with what we implemented in PR #322. We have two functions for application interface. `func (a *applicationImpl) RawResources() (ResourceCollection, error)`  and `func (a *applicationImpl) Resources() (ResourceCollection, error)`. Both `RawResources` and `Resources` have recursive structure. 

Here `NewFromManifest` deals with only one manifest object. It doesn't have any recursion.

pkg/kinflate/resource/resourcecollection.go
mengqiy: > type ResourceCollection types.KObject

This is a intermediate state. After some cleanup in a followup PR, KObject will be gone and only ResourceCollection will be left.

apelisse: Oh, great!

pkg/kinflate/resource/manifestresource.go
apelisse: I think it makes sense, but it's still confusing. I would expect that "NewFromManifest" function, that loads resources from the manifest should load all the resources from the manifest? This is kind of triaging in a confusing way?

pkg/kinflate/resource/resourcecollection.go
mengqiy: I think this line should be moved to the types pkg which will make it closer to its final state.

pkg/kinflate/resource/manifestresource.go
mengqiy: > I'm wondering if this should be a method of the Application abstraction? 

No, I think it is too concrete to be in the application abstraction.

> i.e. I'm not sure people should use this function directly, should they?

Not sure. But we expose `NewFromPath`, `NewFromConfigMap` and `NewFromSecretGenerator`

The action reasonable to me is that move this to the application pkg and make it private.
Application ~= Manifest

Liujingfang1: I see. It is the function name that causes confusion. Do you have some other name in your mind we can replace `NewFromManifest`?

apelisse: Or maybe we should remove all the references to the `v1alpha1`/Manifest type in Resources, and keep these hidden behind the application?

apelisse: > The action reasonable to me is that move this to the application pkg and make it private.
Application ~= Manifest

Yeah. Or a mix of both:

> we should remove all the references to the v1alpha1/Manifest type in Resources, and keep these hidden behind the application?

As in, we still have things to create resources, but not directly from the Manifest types/sub-types

pkg/kinflate/resource/resourcecollection.go
Liujingfang1: We can add a comment on `ResourceCollection` to explain why we need this type.

pkg/kinflate/resource/manifestresource.go
droot: I think this should be implemented as a private method at the ApplicationImpl struct.

pkg/kinflate/resource/resourcecollection.go
```diff
@@ -0,0 +1,41 @@
...
+
+// NewRCFromManifest returns a ResourceCollection given a manifest object
+func NewRCFromManifest(l loader.Loader, m *manifest.Manifest) (ResourceCollection, error) {
```
droot: Not happy with functions in resource package having knowledge of manifest.

apelisse: Yeah, I'm getting more convinced that we should remove the entire `manifest` package from the resource package.

mengqiy:
PTAL

pkg/kinflate/resource/appresource.go
```diff
@@ -21,8 +21,7 @@ import (
 	"k8s.io/kubectl/pkg/loader"
 )
 
-//  NewFromPath returns a Resource list given a resource path from manifest file.
-func NewFromPath(path string, loader loader.Loader) ([]*Resource, error) {
+func newFromPath(loader loader.Loader, path string) ([]*Resource, error) {
```
droot: Can we call this `resourcesFromPath` ? since this is internal it will not be invoked with `resource.` prefix so newFromPath reads very strange to me.

pkg/kinflate/resource/secret_test.go
```diff
@@ -26,21 +26,19 @@ import (
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
 )
 
-func TestNewFromSecretGenerator(t *testing.T) {
-	secret := manifest.SecretGenerator{
+func makeSecretGenerator() manifest.SecretGenerator {
```
apelisse: :-/

pkg/kinflate/resource/appresource.go
droot: nit: can we call pathList ---> paths ?

apelisse:
Yeah, that's definitely less controversial :D

pkg/kinflate/resource/secret_test.go
droot: I think previous version is better. Less logic in the test code.

pkg/kinflate/resource/secret_test.go
```diff
@@ -55,8 +53,21 @@ func TestNewFromSecretGenerator(t *testing.T) {
 			},
 		},
 	}
+}
+
+func TestNewFromSecretGeneratorList(t *testing.T) {
```
droot: List --> s ?

mengqiy:
PTAL. Should be good to go.

pkg/kinflate/resource/appresource_test.go
```diff
@@ -21,7 +21,7 @@ func makeUnconstructed(name string) *unstructured.Unstructured {
 	}
 }
 
-func TestNewFromPath(t *testing.T) {
+func TestNewFromPaths(t *testing.T) {
```
apelisse: Needs update?

pkg/kinflate/resource/configmap.go
apelisse: Wait, you change one function but not the others?!

mengqiy:
@apelisse Is there a way to see the godoc for a PR?

apelisse:
I really really wish we could. That's a project I'd like to see happen, but didn't have time to think about it so far. I've had email with Dmitri Shuralyov (@shurcooL) who already thought about [this](https://groups.google.com/forum/#!topic/golang-dev/C75U43kkumw). I'm not sure we completely agree on how to do it though :-)

The best way for now is to checkout the pull-request, and locally run godoc. Here is how I usually do that:
```
go get golang.org/x/tools/cmd/godoc 
$GOPATH/bin/godoc -http=:8080 >/dev/null 2>&1 & disown # I let this live forever
git fetch $remote pull/$prnumber/head:pull/$prnumber
git checkout pull/$prnumber
# Now browse localhost:8080/pkg/$pkg
```

mengqiy:
Any other comments?

PR #333, opened 2018-03-02
Change inflate to use Application interface

Liujingfang1:
The change is covered by existing tests.

Note:
We will have a different PR to remove the mode flag and add tests for `kinflate diff`.
The type `KObject` will be removed in the future.

### DIFF
--- a/pkg/kinflate/commands/inflate.go
+++ b/pkg/kinflate/commands/inflate.go
@@ -20,13 +20,16 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 
 	"github.com/spf13/cobra"
 
+	"k8s.io/kubectl/pkg/kinflate/app"
 	"k8s.io/kubectl/pkg/kinflate/tree"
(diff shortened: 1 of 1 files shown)


pkg/kinflate/commands/inflate.go
```diff
@@ -83,20 +86,35 @@ func (o *inflateOptions) Complete(cmd *cobra.Command, args []string) error {
...
-	// Do the transformation for the tree.
-	err = root.Inflate(o.mode)
+	application, err := app.New(rootLoader)
+	if err != nil {
+		return nil, err
+	}
+
+	allResources := types.ResourceCollection{}
+	switch o.mode {
+	case tree.ModeNormal:
```
apelisse: Yeah, I think we should handle diff differently, but sure for now.

Liujingfang1: I have another change for removing `o.mode`. Then `diff` and `inflate` will be handled separately.

PR #334, opened 2018-03-02
Update inflate/diff and corresponding tests

Liujingfang1:
Current code uses a mode flag to switch between noop mode and normal mode when running inflate and diff. As a result, diff.go refers to some function in inflate.go. This change includes
- Rewrite `RunDiff` to call Application interface
- Remove the `mode` flag from both diff.go and inflate.go
- Add tests for `kinflate diff` using the same test data for `kinflate inflate`.


pkg/kinflate/commands/diff.go
```diff
@@ -81,23 +84,38 @@ func (o *diffOptions) RunDiff(out, errOut io.Writer, fs fs.FileSystem) error {
 		Stderr: errOut,
 	}
 
-	inflateOp := inflateOptions{manifestPath: o.manifestPath, mode: tree.ModeNormal}
```
apelisse: This concept of `ModeNormal` and `ModeNoop` should disappear altogether

pkg/kinflate/commands/diff_test.go
```diff
@@ -0,0 +1,119 @@
...
+	ExpectedDiff string `yaml:"expectedDiff"`
+}
+
+func TestDiff(t *testing.T) {
+	const updateEnvVar = "UPDATE_KINFLATE_EXPECTED_DATA"
+	updateKinflateExpected := os.Getenv(updateEnvVar) == "true"
+
+	noopDir, _ := regexp.Compile(`/tmp/noop-[0-9]*/`)
+	transformedDir, _ := regexp.Compile(`/tmp/transformed-[0-9]*/`)
+	timestamp, _ := regexp.Compile(`[0-9]{4}-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1]) (2[0-3]|[01][0-9]):[0-5][0-9]:[0-5][0-9].[0-9]* [+-]{1}[0-9]{4}`)
```
apelisse: testing the output of diff makes me a little sad.

pkg/kinflate/commands/diff_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	fs := fs.MakeRealFS()
+
+	var (
+		name     string
+		testcase DiffTestCase
```
apelisse: why do you create this outside the loop where it's used?

pkg/kinflate/commands/diff_test.go
```diff
@@ -0,0 +1,119 @@
...
+			err = diffOps.RunDiff(buf, os.Stderr, fs)
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+			actualString := string(buf.Bytes())
+			actualString = noopDir.ReplaceAllString(actualString, "/tmp/noop/")
+			actualString = transformedDir.ReplaceAllString(actualString, "/tmp/transformed/")
+			actualString = timestamp.ReplaceAllString(actualString, "YYYY-MM-DD HH:MM:SS")
+			actualBytes := []byte(actualString)
+			if !updateKinflateExpected {
```
apelisse: All of this is kind of sad. Can we replace the diff binary with "true" for the test, and just check that the files are what we expect?

Liujingfang1: The files are deleted once `RunDiff` exits. So in testing, we don't have those files available.

pkg/kinflate/commands/diff.go
Liujingfang1: I agree with you and don't worry about this part. `ModeNormal` and `ModeNoop` are defined in package tree. This PR has removed dependency on package tree from `diff` and `inflate`. We only have three other files in package commands that depend on package tree. Once we refactor `add resource`, `set name-prefix` and `add configmap` commands, we can safely remove package tree.

apelisse:
Tests are good this way.

PR #341, opened 2018-03-06
Add util functions to read/write manifest file and update sub commands

Liujingfang1:
This change removes all the dependency on package tree from sub commands add resource, add configmap, set nameprefix

### DIFF
--- a/pkg/kinflate/commands/util.go
+++ b/pkg/kinflate/commands/util.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package commands
+
+import (
+	"errors"
+	"fmt"
+	"path"
+	"strings"
+
+	"github.com/ghodss/yaml"
+
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	"k8s.io/kubectl/pkg/kinflate/constants"
+	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
+	"k8s.io/kubectl/pkg/kinflate/util/fs"
+)
+
+type manifestFile struct {
+	mPath string
+	fsys  fs.FileSystem
+}
+
+func newManifestFile(mPath string, fsys fs.FileSystem) (*manifestFile, error) {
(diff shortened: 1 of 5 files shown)


pkg/kinflate/commands/util.go
```diff
@@ -0,0 +1,80 @@
...
+	"strings"
+
+	"github.com/ghodss/yaml"
+
+	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
+	"k8s.io/kubectl/pkg/kinflate/constants"
+	"k8s.io/kubectl/pkg/kinflate/util/fs"
+)
+
+func validateManifestFile(mPath string, fsys fs.FileSystem) (string, error) {
```
mengqiy: I'd be happier if you make it Object-orientated by making FileSystem a member of the struct.
@apelisse WDYT?

pkg/kinflate/commands/util.go
```diff
@@ -0,0 +1,80 @@
...
+	f, err := fsys.Stat(mPath)
+	if err != nil {
+		errorMsg := fmt.Sprintf("Manifest (%s) missing\nRun `kinflate init` first", mPath)
+		return "", errors.New(errorMsg)
```
apelisse: You can use `fmt.Errorf` instead? (and everywhere else)

pkg/kinflate/commands/util.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if f.IsDir() {
+		mPath = path.Join(mPath, constants.KubeManifestFileName)
+		_, err = fsys.Stat(mPath)
+		if err != nil {
```
apelisse: what if `mpath` is still a dir at this point? :D

pkg/kinflate/commands/util.go
Liujingfang1: I'm OK with using a struct. How about 
```
type ManifestFile struct {
  mpath string
  fsys    fs.FileSystem
}

func New(path string, fsys fs.FileSystem) ManifestFile  //replace current validateManifestFile
func (ManifestFile) Read() (*Manifest, error)
func (ManifestFile) Write(m *Manifest) error
```

pkg/kinflate/commands/util.go
Liujingfang1: Yes, that's true. The reason of keeping it this way is for future work on backlog item. We want to use the internal error package that Sean and I created previously. 

Actually, I can change this error to use manifest error.

pkg/kinflate/commands/util.go
Liujingfang1: By the join statement with `constants.KubeManifestFileName`, mPath should be a file path.

pkg/kinflate/commands/util.go
Liujingfang1: Actually, this struct and member functions should be private.

pkg/kinflate/commands/util.go
apelisse: Well, unless you have a dir named `Kube-manifest.yaml`. Though unlikely, I agree ;-)

Liujingfang1:
Add `manifestFile` struct and functions
```
newManifestfile()
validate()
read()
write()
```

PR #344, opened 2018-03-08
appending name hash only applies to current layer

mengqiy:
@Liujingfang1 observed a bug the hash can be appended in multiple layers, e.g. `configmap-<hash1>-<hash2>`

The name hash transformer should be only applied to the configmap and secrets generated from the current layer.


Liujingfang1:
Looks good to me. Thank you for the quick fix :)

PR #348, opened 2018-03-09
Use passed-in context in pluginutils config initializer

jberkhahn:
This will grab the context out of the config file if one is specified by the KUBECTL_GLOBAL_FLAG_CONTEXT passed in from the plugin framework.

@carolynvs might want to take a look at this.


mengqiy:
/assign @droot 
Assign to you since you have most context.

jberkhahn:
@droot Any update on this? Need this to be merged so we can switch to using this library in svcat.

PR #350, opened 2018-03-12
cleanup Resouce slice related code

mengqiy:
Change function signature to return ResourceCollection instead of []*Resource.
Cleanup tests and embed test data according the tott article `remove logic from the test`.
Moved ResourceCollection to resource pkg.

Most of this PR is mechanical change, nothing very interesting.

### DIFF
--- a/pkg/kinflate/app/application.go
+++ b/pkg/kinflate/app/application.go
@@ -17,8 +17,6 @@ limitations under the License.
 package app
 
 import (
-	"fmt"
-
 	"github.com/ghodss/yaml"
 
 	manifest "k8s.io/kubectl/pkg/apis/manifest/v1alpha1"
@@ -26,16 +24,15 @@ import (
 	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
 	"k8s.io/kubectl/pkg/kinflate/resource"
 	"k8s.io/kubectl/pkg/kinflate/transformers"
-	"k8s.io/kubectl/pkg/kinflate/types"
 	"k8s.io/kubectl/pkg/loader"
 )
 
 type Application interface {
 	// Resources computes and returns the resources for the app.
-	Resources() (types.ResourceCollection, error)
+	Resources() (resource.ResourceCollection, error)
 	// RawResources computes and returns the raw resources from the manifest.
 	// It contains resources from 1) untransformed resources from current manifest 2) transformed resources from sub packages
-	RawResources() (types.ResourceCollection, error)
+	RawResources() (resource.ResourceCollection, error)
 }
 
 var _ Application = &applicationImpl{}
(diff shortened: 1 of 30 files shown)


mengqiy:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kubectl/350)

pkg/kinflate/app/application.go
```diff
@@ -26,16 +26,15 @@ import (
 	interror "k8s.io/kubectl/pkg/kinflate/internal/error"
 	"k8s.io/kubectl/pkg/kinflate/resource"
 	"k8s.io/kubectl/pkg/kinflate/transformers"
-	"k8s.io/kubectl/pkg/kinflate/types"
 	"k8s.io/kubectl/pkg/loader"
 )
 
 type Application interface {
 	// Resources computes and returns the resources for the app.
-	Resources() (types.ResourceCollection, error)
```
droot: Not a deal breaker but was wondering if we should call it just Collection :) it will then read: `resource.Collection` ?

pkg/kinflate/app/application.go
```diff
@@ -210,14 +198,14 @@ func (a *applicationImpl) getTransformer(patches types.ResourceCollection) (tran
 	return transformers.NewMultiTransformer(ts), nil
 }
 
-func resourceCollectionFromResources(res []*resource.Resource) (types.ResourceCollection, error) {
-	out := types.ResourceCollection{}
+func resourceCollectionFromResources(res []*resource.Resource) (resource.ResourceCollection, error) {
```
Liujingfang1: You can remove this function since the resource pkg already handles this part.

pkg/kinflate/resource/appresource.go
```diff
@@ -17,37 +17,27 @@ limitations under the License.
...
-		res = append(res, &Resource{Data: obj})
-	}
-	return res, nil
+	return decodeToResourceCollection(content)
 }
 
 //  NewFromPaths returns a slice of Resources given a resource path slice from manifest file.
-func NewFromPaths(loader loader.Loader, paths []string) ([]*Resource, error) {
-	allResources := []*Resource{}
+func NewFromPaths(loader loader.Loader, paths []string) (ResourceCollection, error) {
```
droot: I remember some discussion about keeping them as slices, but don't remember the why now :)

pkg/kinflate/app/application.go
mengqiy: `Collection` sounds like an interface, and `ResourceCollection` sounds like a concrete type.

pkg/kinflate/resource/appresource.go
Liujingfang1: We wanted to get rid of `KObject` at first so we kept it as slices. When working on the code for Application interfaces, we realized that we don't achieve it without a map from GVKN to unstructured. So we plan to rename `KObject` to `ResourceCollection`.

mengqiy:
I guess this PR is OK to merge. We can solve the tech debate later. 
Can someone approve it?

PR #352, opened 2018-03-12
fix bug when update name ref in multi layer setup

mengqiy:


pkg/kinflate/app/application.go
```diff
@@ -95,16 +95,32 @@ func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	t, err := a.getTransformer(patches)
 	if err != nil {
 		return nil, err
 	}
-	err = t.Transform(allRes)
+	err = t.Transform(allTransRes)
```
Liujingfang1: I'm confused at the end of this function. You created `allRes` and performed overly transformation on it. You also created `allTransRes` and performed other transformation on it. At the end, I expect to see some kind of merging or combining logic.

mengqiy: Added some comments to explain it.
PTAL.

Liujingfang1: Now I understand it. Looks good to me. I also verified the change against examples. Thank you for this fix.

PR #354, opened 2018-03-14
unstructpath: Create json path type package.

apelisse:
This package let's you find specific path in your unstruct types, kind
of like jsonpath does. But this is statically compile.

This replaces #258 

It doesn't have the controversial types anymore.

### DIFF
--- a/pkg/framework/unstructpath/example_test.go
+++ b/pkg/framework/unstructpath/example_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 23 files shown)


pkg/framework/unstructpath/example_test.go
```diff
@@ -0,0 +1,90 @@
...
+	    }
+	  }
+	}`
+	u := map[string]interface{}{}
+	if err := yaml.Unmarshal([]byte(data), &u); err != nil {
+		panic(err)
+	}
+
+	// The authors of all books in the store. Returns a list of strings.
+	Children().Map().Field("book").Children().Map().Field("author").String().SelectFrom(u)
```
pwittrock:
```go
From(Children().Field("book").Field("author").String())
```

PR #360, opened 2018-03-19
predicates: Rename Value into Interface

apelisse:
Since the original "Value" class is gone, it doesn't make sense anymore
to refer to that name. Now it's simply an "interface{}", so the name
Interface is more suited.


apelisse:
Mechanical again, no other change than `s/Value/Interface/`

apelisse:
I've only converted capitalized `Value`, I'll do a second pass for `value`

apelisse:
Fixed.

PR #368, opened 2018-03-22
updated namereference config for missing references.

droot:
highlights:
RBAC references
ServiceAccountName reference
StatefulSet headless service reference

### DIFF
--- a/pkg/kustomize/transformers/namereferenceconfig.go
+++ b/pkg/kustomize/transformers/namereferenceconfig.go
@@ -339,4 +339,134 @@ var defaultNameReferencePathConfigs = []referencePathConfig{
 			},
 		},
 	},
+	{
+		// StatefulSet references headless service, so need to update the references.
+		referencedGVK: schema.GroupVersionKind{
+			Version: "v1",
+			Kind:    "Service",
+		},
+		pathConfigs: []PathConfig{
+			{
+				GroupVersionKind: &schema.GroupVersionKind{
+					Group: "apps",
+					Kind:  "StatefulSet",
+				},
+				Path:               []string{"spec", "serviceName"},
+				CreateIfNotPresent: false,
+			},
+		},
+	},
+	{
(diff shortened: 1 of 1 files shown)


pkg/kinflate/transformers/namereferenceconfig.go
```diff
@@ -339,4 +339,109 @@ var defaultNameReferencePathConfigs = []referencePathConfig{
...
+					Group: "rbac.authorization.k8s.io",
+					Kind:  "RoleBinding",
+				},
+				Path:               []string{"roleRef", "name"},
+				CreateIfNotPresent: false,
+			},
+			{
+				GroupVersionKind: &schema.GroupVersionKind{
+					Group: "rbac.authorization.k8s.io",
+					Kind:  "ClusterRoleBinding",
```
mengqiy: IIRC `ClusterRoleBinding` can't ref another `Role`.
I think we should drop this entry.

pkg/kinflate/transformers/namereferenceconfig.go
```diff
@@ -339,4 +339,109 @@ var defaultNameReferencePathConfigs = []referencePathConfig{
...
+				GroupVersionKind: &schema.GroupVersionKind{
+					Group: "apps",
+					Kind:  "StatefulSet",
+				},
+				Path:               []string{"spec", "template", "spec", "serviceAccountName"},
+				CreateIfNotPresent: false,
+			},
+			{
+				GroupVersionKind: &schema.GroupVersionKind{
+					Group: "apps",
```
mengqiy: We can drop the `Group` here so it can also match `Deployment` in the extensions group.

I'd like to see you add config for all workload: Pod, RC, ReplicaSet, DaemonSet, Job, CronJob :)
Note: DaemonSet are in both `apps` and `extensions` group, so you don't need to specify the `Group` for it.

PR #370, opened 2018-03-22
support multiple patches for the same GVKN

mengqiy:
Before: multiple patches against the same GVKN is not supported, because loaded patches are stored in the ResourceCollection which is indexed by GVKN. So it will complain if there are multiple patches for one GVKN.

After: loaded patches are stored in a slice of Resource.

### DIFF
--- a/pkg/kinflate/resource/appresource.go
+++ b/pkg/kinflate/resource/appresource.go
@@ -20,24 +20,38 @@ import (
 	"k8s.io/kubectl/pkg/loader"
 )
 
-func resourcesFromPath(loader loader.Loader, path string) (ResourceCollection, error) {
-	content, err := loader.Load(path)
-	if err != nil {
-		return nil, err
-	}
-
-	return decodeToResourceCollection(content)
-}
-
-//  NewFromPaths returns a slice of Resources given a resource path slice from manifest file.
-func NewFromPaths(loader loader.Loader, paths []string) (ResourceCollection, error) {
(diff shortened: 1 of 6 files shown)


pkg/kinflate/resource/appresource.go
```diff
@@ -20,24 +20,38 @@ import (
...
 		if err != nil {
 			return nil, err
 		}
 		allResources = append(allResources, res)
 	}
 	return Merge(allResources...)
 }
+
+//  NewFromPatches returns a slice of Resources given a patch path slice from manifest file.
+func NewFromPatches(loader loader.Loader, paths []string) ([]*Resource, error) {
```
Liujingfang1: The function `NewFromResources` and `NewFromPatches` have the same input arguments, similar function logic and different return type. I suggest to remove `NewFromPatches` and keep previous `NewFromPaths`. Add another function `ExtractFromResourceCollection` which can give `[]*Resource`.

mengqiy: What you suggest will not always work.
If we have multiple patches for the same GVKN, `NewFromResources` will fail since it tries to return a `ResourceCollection` indexed by GVKN.
IOW, this flow works: `loader` -> `[]Resource` -> `ResourceCollection`
this don't: `loader` -> `ResourceCollection` -> `[]Resource`

Liujingfang1: yes, you are right. I forgot about it when I was reviewing the code. 

LGTM!

PR #373, opened 2018-03-23
Add different behavior for configmaps in overlay

Liujingfang1:
In our current kindflate code, we don't support redefining configmap or secrets in overlays. If we want to change a configmap in the base, we are only allowed to apply a patch file containing the new configmap content.

This change is to enable us to redefine configmap in overlays. For example, in `base/Kube-manifest.yaml`, we define
```configMaps:
- name: demo-configmap
  files:
    - application.properties
```
In `instance/staging/Kube-manifest.yaml`, if we define it as
```configMaps:
- name: demo-configmap
  behavior: replace
  files:
    - staging-application.properties
```
Then `staging-application.properties` is going to be used instead of `application.properties`.

In `instance/production/Kube-manifest.yaml`, if we define it as
```configMaps:
- name: demo-configmap
  behavior: merge
  files:
    - production-env.properties
```
Then both `application.properties` and `production-env.properties` are used

### DIFF
--- a/pkg/apis/manifest/v1alpha1/types.go
+++ b/pkg/apis/manifest/v1alpha1/types.go
@@ -132,6 +132,12 @@ type ConfigMap struct {
 	// hash(content of configmap).
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 
+	// Behavior of the configmap, must be one of create, merge and replace
+	// 'create': create a new one;
+	// 'replace': replace the existing one;
+	// 'merge': merge the existing one.
+	Behavior string `json:"behavior,omitempty" yaml:"behavior,omitempty"`
+
 	// DataSources for configmap.
 	DataSources `json:",inline,omitempty" yaml:",inline,omitempty"`
 }
--- a/pkg/kinflate/app/application.go
+++ b/pkg/kinflate/app/application.go
@@ -33,6 +33,8 @@ import (
 type Application interface {
 	// Resources computes and returns the resources for the app.
 	Resources() (resource.ResourceCollection, error)
+	// SemiResources computes and returns the resources without name hash and name reference for the app
+	SemiResources() (resource.ResourceCollection, error)
 	// RawResources computes and returns the raw resources from the manifest.
 	// It contains resources from 1) untransformed resources from current manifest 2) transformed resources from sub packages
 	RawResources() (resource.ResourceCollection, error)
@@ -63,9 +65,28 @@ func New(loader loader.Loader) (Application, error) {
 }
 
 // Resources computes and returns the resources from the manifest.
+// The namehashing for configmap/secrets and resolving name reference is only done
+// in the most top overlay once at the end of getting resources.
 func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
+	res, err := a.SemiResources()
+	if err != nil {
+		return nil, err
+	}
+	t, err := a.getHashAndReferenceTransformer()
+	if err != nil {
+		return nil, err
+	}
+	err = t.Transform(res)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+// SemiResources computes and returns the resources without name hash and name reference for the app
+func (a *applicationImpl) SemiResources() (resource.ResourceCollection, error) {
 	errs := &interror.ManifestErrors{}
-	raw, err := a.RawResources()
+	raw, err := a.rawResources()
 	if err != nil {
 		errs.Append(err)
 	}
@@ -82,9 +103,8 @@ func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
 	if err != nil {
 		return nil, err
 	}
-	// Only append hash for generated configmaps and secrets.
-	nht := transformers.NewNameHashTransformer()
-	err = nht.Transform(res)
+
+	allRes, err := resource.MergeWithOverride(raw, res)
 	if err != nil {
 		return nil, err
 	}
@@ -98,37 +118,11 @@ func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
 		return nil, errs
 	}
 
-	// Reindex the Raw Resources (resources from sub package and resources field of this package).
-	// raw is a ResourceCollection (map) from <GVK, original name> to object with the transformed name.
-	// transRaw is a ResourceCollection (map) from <GVK, transformed name> to object with the transformed name.
-	transRaw := reindexResourceCollection(raw)
-	// allRes contains the resources that are indexed by the original names (old names).
-	// allTransRes contains the resources that are indexed by the transformed names (new names).
-	// allRes and allTransRes point to the same set of objects with new names.
-	allTransRes, err := resource.Merge(res, transRaw)
+	t, err := a.getTransformer(patches)
 	if err != nil {
 		return nil, err
 	}
-	allRes, err := resource.Merge(res, raw)
-	if err != nil {
-		return nil, err
-	}
-
-	ot, err := transformers.NewOverlayTransformer(patches)
-	if err != nil {
-		return nil, err
-	}
-	// Overlay transformer uses the ResourceCollection indexed by the original names.
-	err = ot.Transform(allRes)
-	if err != nil {
-		return nil, err
-	}
-
-	t, err := a.getTransformer()
-	if err != nil {
-		return nil, err
-	}
-	err = t.Transform(allTransRes)
+	err = t.Transform(allRes)
 	if err != nil {
 		return nil, err
 	}
@@ -137,7 +131,25 @@ func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
 }
 
 // RawResources computes and returns the raw resources from the manifest.
+// The namehashing for configmap/secrets and resolving name reference is only done
+// in the most top overlay once at the end of getting resources.
 func (a *applicationImpl) RawResources() (resource.ResourceCollection, error) {
+	res, err := a.rawResources()
+	if err != nil {
+		return nil, err
+	}
+	t, err := a.getHashAndReferenceTransformer()
+	if err != nil {
+		return nil, err
+	}
+	err = t.Transform(res)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
+
+func (a *applicationImpl) rawResources() (resource.ResourceCollection, error) {
 	subAppResources, errs := a.subAppResources()
 	resources, err := resource.NewFromResources(a.loader, a.manifest.Resources)
 	if err != nil {
@@ -166,7 +178,7 @@ func (a *applicationImpl) subAppResources() (resource.ResourceCollection, *inter
 			continue
 		}
 		// Gather all transformed resources from subpackages.
-		subAppResources, err := subapp.Resources()
+		subAppResources, err := subapp.SemiResources()
 		if err != nil {
 			errs.Append(err)
 			continue
@@ -181,12 +193,18 @@ func (a *applicationImpl) subAppResources() (resource.ResourceCollection, *inter
 }
 
 // getTransformer generates the following transformers:
-// 1) name prefix
-// 2) apply labels
-// 3) apply annotations
-// 4) update name reference
-func (a *applicationImpl) getTransformer() (transformers.Transformer, error) {
+// 1) apply overlay
+// 2) name prefix
+// 3) apply labels
+// 4) apply annotations
+func (a *applicationImpl) getTransformer(patches []*resource.Resource) (transformers.Transformer, error) {
 	ts := []transformers.Transformer{}
+
+	ot, err := transformers.NewOverlayTransformer(patches)
+	if err != nil {
+		return nil, err
+	}
+	ts = append(ts, ot)
 
 	npt, err := transformers.NewDefaultingNamePrefixTransformer(string(a.manifest.NamePrefix))
 	if err != nil {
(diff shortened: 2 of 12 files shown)


jordanwilson230:
@Liujingfang1 For those looking for a way to manage their k8 workloads as they grow in size and complexity, this is a helpful feature.  To be frank, I don't see much benefit of kinflate without it supporting this. Anyways, thank you for the contribution.

Liujingfang1:
@jordanwilson230 Thank you for your comments. Today our team had a discussion about how the overlay should behave for configmaps. We'd like to hear your thoughts as well, especially for redefining configmaps in overlay, do you want the configs to be replaced or merged? What's the usage you can think of. Any idea or comment is appreciated.

jordanwilson230:
@Liujingfang1 My opinions seem to shift every now and then regarding your question :)  I'd probably opt for a merging strategy, so I assume some way to define the top-level directory path within the Kube-manifest.yaml?

This is outside the scope of your (awesome :) work here, but aside from overlays, an ideal method for handling the variation in the namespaces/environments might be to abstract configmap/secret configuration the same way that kubectl and gcloud abstract cluster information, context, etc.  We certainly tinker with and edit our configmaps, but aside from having to deal with the per-namespace variation, they are actually quite static.  So, in short, `kubectl config use-context dev` automatically inherits the preset configurations for that context (similar command for applying changes). In any case, this is just my personal opinion, and I don't want to get too far out of scope of the PR here!

One last reason I'd prefer the merged approach, is simply that it does something I could not perform myself.  Let me know if I can be of any other help, and thank you again for your work!

Liujingfang1:
I updated the code according to the doc. Please take another look.

jordanwilson230:
@Liujingfang1 The best of both 👍  Excited and hoping to see this merge soon.

pkg/kinflate/app/application.go
```diff
@@ -30,6 +30,8 @@ import (
 type Application interface {
 	// Resources computes and returns the resources for the app.
 	Resources() (resource.ResourceCollection, error)
+	// SemiResources computes and returns the resources without name hash and name reference for the app
+	SemiResources() (resource.ResourceCollection, error)
```
mengqiy: This doesn't have to be public.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,61 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+	for _, m := range maps[1:] {
+		for key, value := range m {
+			mergedMap[key] = value
+		}
+	}
+	return mergedMap
+}
+
+func mergeData(mergedTo map[string]interface{}, mergedFrom map[string]interface{}) {
+	mergedToMap, toOK := mergedTo["data"].(map[string]interface{})
```
mengqiy: We should try to avoid using a hardcoded string key here. It should  be possible. :)

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,61 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+				all[gvkn] = obj
+				obj.Data.GetAnnotations()
+			}
+		}
+	}
+	return all, nil
+}
+
+func mergeMap(maps ...map[string]string) map[string]string {
+	mergedMap := maps[0]
```
mengqiy: Why not just create a new map for better readability?

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,61 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+					all[gvkn].Data.SetLabels(mergeMap(obj.Data.GetLabels(), all[gvkn].Data.GetLabels()))
+					all[gvkn].Data.SetAnnotations(mergeMap(obj.Data.GetAnnotations(), all[gvkn].Data.GetAnnotations()))
+					all[gvkn].Data.SetName(obj.Data.GetName())
+					mergeData(all[gvkn].Data.Object, obj.Data.Object)
+				default:
+					return nil, fmt.Errorf("The behavior of %#v must be one of create, merge and replace", gvkn)
+				}
+			} else {
+				all[gvkn] = obj
+				obj.Data.GetAnnotations()
```
mengqiy: It seems this line is not used anywhere.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,61 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+				case "merge":
+					all[gvkn].Data.SetLabels(mergeMap(obj.Data.GetLabels(), all[gvkn].Data.GetLabels()))
+					all[gvkn].Data.SetAnnotations(mergeMap(obj.Data.GetAnnotations(), all[gvkn].Data.GetAnnotations()))
+					all[gvkn].Data.SetName(obj.Data.GetName())
+					mergeData(all[gvkn].Data.Object, obj.Data.Object)
+				default:
+					return nil, fmt.Errorf("The behavior of %#v must be one of create, merge and replace", gvkn)
+				}
+			} else {
+				all[gvkn] = obj
```
mengqiy: IMO when it's `replace` it doesn't appear, we should error out.

pkg/kinflate/app/application.go
Liujingfang1: `SemiResources` is called by the sub applications. They are not available if set to private.

pkg/kinflate/resource/util.go
Liujingfang1: You are right. We don't need this line.

pkg/kinflate/resource/util.go
Liujingfang1: yes, we should avoid this. Let's discuss offline.

pkg/kinflate/resource/util.go
Liujingfang1: Good point. That's one error I listed in the doc but forgot to add it here. Thank you for catching it.

Liujingfang1: This is not easy case to deal with. Since we have both `allTransRes` and `allRes` and merge will take action for one of them, but not both.

Liujingfang1:
Remove the  reindexing part of resourcecollection. The reason we added reindexing was to fix the bug of name reference in the base. https://github.com/kubernetes/kubectl/pull/352  

In this change, we don't do name reference until the last overlay. So in the base layers, the name reference is still with the original name. No need to reindex.

pkg/kinflate/app/application.go
```diff
@@ -30,6 +30,8 @@ import (
 type Application interface {
 	// Resources computes and returns the resources for the app.
 	Resources() (resource.ResourceCollection, error)
+	// semiResources computes and returns the resources without name hash and name reference for the app
+	semiResources() (resource.ResourceCollection, error)
```
mengqiy: Not a big fun of this naming. But I can't find a better one and this is private.
So I guess we can leave it as is. And see if someone in the future can find a better one.

pkg/kinflate/app/application.go
```diff
@@ -61,8 +63,25 @@ func New(loader loader.Loader) (Application, error) {
 
 // Resources computes and returns the resources from the manifest.
 func (a *applicationImpl) Resources() (resource.ResourceCollection, error) {
```
mengqiy: Please add a comment explaining name hashing and name ref are only updated at the very end.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,71 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
 
 	return all, nil
 }
+
+// Merge will merge all of the entries in the slice of ResourceCollection with Override
+// If there is already an entry with the same GVKN exists, different actions are performed according to value of Behavior field
+// 'create': create a new one;
```
mengqiy: I think it's valuable to add this to the `types.go` file.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,71 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+					obj.Data.SetAnnotations(mergeMap(originalObj.Data.GetAnnotations(), obj.Data.GetAnnotations()))
+					obj.Data.SetName(originalObj.Data.GetName())
+					mergeData(obj.Data.Object, originalObj.Data.Object, obj.Data.Object)
+					all[gvkn] = obj
+				default:
+					return nil, fmt.Errorf("The behavior of %#v must be one of create, merge and replace", gvkn)
+				}
+			} else {
+				switch obj.Behavior {
+				case "", "create", "replace":
```
mengqiy: > Replace non existing configmaps

From the spec doc you wrote, it seems `replace` should also error out.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +90,71 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+	mergedMap := map[string]string{}
+	for _, m := range maps {
+		for key, value := range m {
+			mergedMap[key] = value
+		}
+	}
+	return mergedMap
+}
+
+func mergeData(mergedTo map[string]interface{}, maps ...map[string]interface{}) {
```
mengqiy: I'd like to call it `mergeConfigMap` which is more descriptive.
And after we sync to new k8s.io/api, we can hangle [BinaryData](https://github.com/kubernetes/kubernetes/blob/3538676841c33c87b8790947b898b2c61bc49c9d/staging/src/k8s.io/api/core/v1/types.go#L5018-L5026) here.

And please leave a TODO somewhere about handling `BinaryData` to ensure we will not forget it. :)

Liujingfang1: Sounds good, will add it.

pkg/kinflate/resource/util.go
Liujingfang1: Will fix it.

pkg/kinflate/constants/constants.go
```diff
@@ -19,3 +19,8 @@ package constants
 // KubeManifestFileName is the Well-Known File Name for a kubernetes app manifest.
 const KubeManifestSuffix = ".yaml"
 const KubeManifestFileName = "Kube-manifest" + KubeManifestSuffix
+
+// Configmap behaviors
+const Create = "create"
+const Replace = "replace"
+const Merge = "merge"
```
droot: We should suffix or prefix the above with `Behavior` or `ConflictBehavior` term.

pkg/kinflate/app/application.go
droot: private functions in an interface ? Haven't seen these before. Looks like a code smell :) to me.

pkg/kinflate/app/application.go
```diff
@@ -63,9 +65,28 @@ func New(loader loader.Loader) (Application, error) {
...
+	t, err := a.getHashAndReferenceTransformer()
+	if err != nil {
+		return nil, err
+	}
+	err = t.Transform(res)
+	if err != nil {
+		return nil, err
+	}
+	return res, nil
+}
```
droot: `semiResources` naming is making things hard to understand. 

I am realizing that now the core logic is not inline with the abstractions we had earlier. Transformers are mixed in a weird way to understand the flow.

I am still trying to form a mental model about the core logic so that it can be simplified. One thing is becoming clear, we need some logic which needs to be applied at the last layer (top most one), for ex. appending hashes to configmap/secrets and then as a result updating name references for those. I hit similar ones while prototyping downward APIs.

pkg/apis/manifest/v1alpha1/types.go
```diff
@@ -132,6 +132,12 @@ type ConfigMap struct {
 	// hash(content of configmap).
 	Name string `json:"name,omitempty" yaml:"name,omitempty"`
 
+	// Behavior of the configmap, must be one of create, merge and replace
+	// 'create': create a new one;
+	// 'replace': replace the existing one;
+	// 'merge': merge the existing one.
+	Behavior string `json:"behavior,omitempty" yaml:"behavior,omitempty"`
```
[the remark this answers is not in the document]
Liujingfang1: I feel the actual usage is different from resolving conflict or tightly related with conflict. It's more related to override. So far our discussion about conflict is mostly about conflicts in one layer, such as conflict from two patches in the same overlay.

pkg/kinflate/resource/util.go
```diff
@@ -90,3 +91,72 @@ func Merge(rcs ...ResourceCollection) (ResourceCollection, error) {
...
+					originalObj := all[gvkn]
+					obj.Data.SetLabels(mergeMap(originalObj.Data.GetLabels(), obj.Data.GetLabels()))
+					obj.Data.SetAnnotations(mergeMap(originalObj.Data.GetAnnotations(), obj.Data.GetAnnotations()))
+					obj.Data.SetName(originalObj.Data.GetName())
+					mergeConfigmap(obj.Data.Object, originalObj.Data.Object, obj.Data.Object)
+					all[gvkn] = obj
+				default:
+					return nil, fmt.Errorf("The behavior of %#v must be one of merge and replace since it already exists in the base", gvkn)
+				}
+			} else {
```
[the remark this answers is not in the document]
Liujingfang1: We can do that. Then the switch case will call those functions directly.

pkg/kinflate/constants/constants.go
Liujingfang1: I will add it. Thank you.

pkg/kinflate/app/application.go
Liujingfang1: You are right. The core logic is changed. The transformers are split and put into different parts. 
Appending hashes and name reference are applied only on the top layer. I also feel that we can make this application interface better and more  clear to understand, but haven't got any concrete ideas.

pkg/kinflate/app/application.go
mengqiy: [Here](https://github.com/kubernetes/kubectl/pull/373/files#diff-b38a7fa92a6de50adb2d8e890664c5fcR181) is using this method. `subapp` is an interface instead of a concrete type.

pkg/kinflate/resource/util.go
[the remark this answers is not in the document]
Liujingfang1: This is done.

Liujingfang1:
change `semiResources` to `SemiResources`.

PR #375, opened 2018-03-23
Conflict detection in kinflate

mengqiy:
Support conflict detection for patches in the same Kube-manifest.yaml.

This depends on https://github.com/kubernetes/kubernetes/pull/61905.

### DIFF
--- a/pkg/kinflate/commands/diff_test.go
+++ b/pkg/kinflate/commands/diff_test.go
@@ -36,7 +36,8 @@ type DiffTestCase struct {
 	Args        []string `yaml:"args"`
 	Filename    string   `yaml:"filename"`
 	// path to the file that contains the expected output
-	ExpectedDiff string `yaml:"expectedDiff"`
+	ExpectedDiff  string `yaml:"expectedDiff"`
+	ExpectedError string `yaml:"expectedError"`
 }
 
 func TestDiff(t *testing.T) {
@@ -89,9 +90,18 @@ func TestDiff(t *testing.T) {
 			}
 			buf := bytes.NewBuffer([]byte{})
 			err = diffOps.RunDiff(buf, os.Stderr, fs)
-			if err != nil {
+			switch {
+			case err != nil && len(testcase.ExpectedError) == 0:
 				t.Errorf("unexpected error: %v", err)
+			case err != nil && len(testcase.ExpectedError) != 0:
+				if !strings.Contains(err.Error(), testcase.ExpectedError) {
+					t.Errorf("expected error to contain %q but got: %v", testcase.ExpectedError, err)
+				}
+				return
+			case err == nil && len(testcase.ExpectedError) != 0:
+				t.Errorf("unexpected no error")
 			}
+
 			actualString := string(buf.Bytes())
 			actualString = noopDir.ReplaceAllString(actualString, "/tmp/noop/")
 			actualString = transformedDir.ReplaceAllString(actualString, "/tmp/transformed/")
--- a/pkg/kinflate/transformers/overlay.go
+++ b/pkg/kinflate/transformers/overlay.go
@@ -22,6 +22,7 @@ import (
 
 	jsonpatch "github.com/evanphx/json-patch"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/kubectl/pkg/kinflate/resource"
@@ -45,8 +46,14 @@ func NewOverlayTransformer(overlay []*resource.Resource) (Transformer, error) {
 
 // Transform apply the overlay on top of the base resources.
 func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollection) error {
+	// Merge and then index the patches by GVKN.
+	overlays, err := o.mergePatches()
+	if err != nil {
+		return err
+	}
+
 	// Strategic merge the resources exist in both base and overlay.
-	for _, overlay := range o.overlay {
+	for _, overlay := range overlays {
 		// Merge overlay with base resource.
 		gvkn := overlay.GVKN()
 		base, found := baseResourceMap[gvkn]
@@ -82,10 +89,14 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
 			// TODO: Change this to use the new Merge package.
 			// Store the name of the base object, because this name may have been munged.
 			// Apply this name to the StrategicMergePatched object.
-			merged, err = strategicpatch.StrategicMergeMapPatch(
-				base.Data.UnstructuredContent(),
-				overlay.Data.UnstructuredContent(),
-				versionedObj)
+			lookupPatchMeta, err := strategicpatch.NewPatchMetaFromStruct(versionedObj)
+			if err != nil {
+				return err
+			}
+			merged, err = strategicpatch.StrategicMergeMapPatchUsingLookupPatchMeta(
+				base.Data.Object,
+				overlay.Data.Object,
+				lookupPatchMeta)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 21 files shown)


pkg/kinflate/commands/diff_test.go
```diff
@@ -89,9 +90,18 @@ func TestDiff(t *testing.T) {
 			}
 			buf := bytes.NewBuffer([]byte{})
 			err = diffOps.RunDiff(buf, os.Stderr, fs)
-			if err != nil {
+			switch {
+			case err != nil && len(testcase.ExpectedError) == 0:
 				t.Errorf("unexpected error: %v", err)
+			case err != nil && len(testcase.ExpectedError) != 0:
+				if !strings.Contains(err.Error(), testcase.ExpectedError) {
```
apelisse: I don't think it's super good practice to grep on error strings :-/

PS: Same below

pkg/kinflate/commands/diff_test.go
```diff
@@ -89,9 +90,18 @@ func TestDiff(t *testing.T) {
...
+					t.Errorf("expected error to contain %q but got: %v", testcase.ExpectedError, err)
+				}
+				return
+			case err == nil && len(testcase.ExpectedError) != 0:
+				t.Errorf("unexpected no error")
 			}
+
 			actualString := string(buf.Bytes())
 			actualString = noopDir.ReplaceAllString(actualString, "/tmp/noop/")
```
apelisse: Not super easy to parallelize etc :-/

pkg/kinflate/transformers/overlay.go
```diff
@@ -59,39 +65,143 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
...
 			// Use JSON merge patch to handle types w/o schema
-			baseBytes, err := json.Marshal(base.Data)
+			merged, err = jsonMergePatch(base.Data.Object, overlay.Data.Object, jsonpatch.MergePatch)
 			if err != nil {
 				return err
 			}
-			patchBytes, err := json.Marshal(overlay.Data)
+		case err != nil:
+			return err
+		default:
```
apelisse: I'm really not convinced that this very weird switch is the most readable. Especially with err being defined in multiple different places :-)

pkg/kinflate/transformers/overlay.go
```diff
@@ -59,39 +65,143 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
...
+		}
+		base.Data.SetName(baseName)
+		baseResourceMap[gvkn].Data.Object = merged
+	}
+	return nil
+}
+
+// mergePatches merge and index patches by GVKN.
+// It errors out if there is conflict between patches.
+func (o *overlayTransformer) mergePatches() (resource.ResourceCollection, error) {
```
apelisse: This function is too long. It's cyclomatic complexity terrifies me.

pkg/kinflate/transformers/overlay.go
```diff
@@ -59,39 +65,143 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
...
+	for _, patch := range o.overlay {
+		gvkn := patch.GVKN()
+		existing, found := rc[gvkn]
+		if !found {
+			rc[gvkn] = patch
+			continue
+		}
+		var merged map[string]interface{}
+		versionedObj, err := scheme.Scheme.New(gvkn.GVK)
+		switch {
```
apelisse: I have the feeling that this switch should be a "virtual table"/dispatch rather than a switch, this would make it much easier to test/understand.

pkg/kinflate/transformers/overlay_test.go
```diff
@@ -149,6 +150,277 @@ func TestOverlayRun(t *testing.T) {
...
+			Name: "deploy1",
+		}: &resource.Resource{
+			Data: &unstructured.Unstructured{
+				Object: map[string]interface{}{
+					"apiVersion": "apps/v1",
+					"kind":       "Deployment",
+					"metadata": map[string]interface{}{
+						"name": "deploy1",
+					},
+					"spec": map[string]interface{}{
```
apelisse: Can we use much simpler examples maybe? Much more to the point?

pkg/kinflate/transformers/overlay.go
```diff
@@ -59,39 +65,143 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
 		switch {
 		case runtime.IsNotRegisteredError(err):
 			// Use JSON merge patch to handle types w/o schema
-			baseBytes, err := json.Marshal(base.Data)
+			merged, err = jsonMergePatch(base.Data.Object, overlay.Data.Object, jsonpatch.MergePatch)
```
Liujingfang1: When merging the patches in the overlay, you called conflict detection and conflict locating functions before `jsonMergePatch`. You may want to do the same here. If there is a conflict, it will fail and the error message may not be very descriptive.

pkg/kinflate/commands/diff_test.go
mengqiy: There are 2 reasons that I choose not to use it.
- the error message doesn't make sense for us: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/mergepatch/errors.go#L88-L91
- all test config  are loaded from a yaml config file. IOW it doesn't know about the error type.

pkg/kinflate/commands/diff_test.go
mengqiy: cc: @Liujingfang1

pkg/kinflate/transformers/overlay.go
mengqiy: Not following.
Can you be more concrete?

pkg/kinflate/transformers/overlay_test.go
mengqiy: Will try to simplify this.

pkg/kinflate/transformers/overlay.go
mengqiy: It sounds like you are talking about detecting conflict when applying patch to base.
That is a new feature that is out of the scope of this PR. It just override for now.

pkg/kinflate/transformers/overlay.go
mengqiy: > I'm really not convinced that this very weird switch is the most readable.

Suggestions?

> Especially with err being defined in multiple different places

Not sure if I understand it.
It is switching on the same err which is returned by `versionedObj, err := scheme.Scheme.New(gvkn.GVK)`.

pkg/kinflate/transformers/overlay.go
mengqiy: Will fix.

pkg/kinflate/transformers/overlay.go
apelisse: OK So let's look at the `case` statement. This one does:
- Check if there are conflicts,
- if there are, locate them and return,
- if we don't, merge the objects.

Now let's look at the default `case` statement. What do you do?
- Check if it has conflicts,
- if there are, locate them and return,
- if we don't, merge the objects.

That logic (check-locate-return-merge) is **orthogonal** to how each of these steps are actually performed (using jsonpath or smp). And that logic is reproduced twice.

I'd rather see something like this:
```
var mergeStrategy MergeStrategyInterface

// The following piece of code decides what type of strategy to use.
if !runtime.IsNotRegisteredError(err) {
   return err
}
if err != nil {
   mergeStrategy = JsonPatch{} // add other useful stuff in there
} else {
  mergeStrategy = SMP{Schema: schema} // add other useful stuff in there
}

// The following piece has the actual logic that we care about:
// If conflict, where are they, otherwise merge:
if mergeStrategy.HasConflict(object1, object2) {
   return mergeStrategy.FindConflict(object1, object2)
} else {
  return mergeStrategy.Merge(object1, object2)
}

...
// The details of how each strategy works come somewhere else...
```

The same applies for the two conflict detection functions below.

pkg/kinflate/transformers/overlay.go
```diff
@@ -59,39 +65,143 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
...
+		gvkn := patch.GVKN()
+		existing, found := rc[gvkn]
+		if !found {
+			rc[gvkn] = patch
+			continue
+		}
+		var merged map[string]interface{}
+		versionedObj, err := scheme.Scheme.New(gvkn.GVK)
+		switch {
+		case runtime.IsNotRegisteredError(err):
```
apelisse: On a completely unrelated note, why do we need to have the types registered to use smp?

pkg/kinflate/transformers/overlay.go
mengqiy: I see. This is similar to what we do in apply v1.5 you developed.

pkg/kinflate/transformers/overlay.go
mengqiy: SMP need the scheme to generate a versioned object for reference.
We can easily switch it to use OpenAPI. I have the code ready, but we need to convince people we need to contact the server.

pkg/kinflate/transformers/overlay_test.go
mengqiy: Simplified it a bit.

apelisse:
Looks better. Thanks. I'll let someone else approve.

pkg/kinflate/transformers/overlay.go
```diff
@@ -95,3 +106,59 @@ func (o *overlayTransformer) Transform(baseResourceMap resource.ResourceCollecti
...
+				return nil, err
+			}
+		}
+
+		conflict, err := cd.hasConflict(existing.Data, patch.Data)
+		if err != nil {
+			return nil, err
+		}
+		if conflict {
+			conflictingPatch, err := cd.findConflict(ix, patches)
```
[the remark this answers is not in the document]
mengqiy: Suppose you have patches A, B, C and D. And A conflicts with D.
After merging A, B and C, you will have (A+B+C) which is `existing.Data`.
Then you find merged patch (A+B+C)  conflicts with D.
You need to use the patches before merging to find out A conflicts with D. So you need to pass in the conflicting patch (it can be identified by `ix`) and all of the patches (`patches`).
IOW, we need n patches instead of 2. `existing.Data` and `patch.Data` are 2 patches, we can not get some useful error from it.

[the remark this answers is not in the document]
Liujingfang1: The explanation helps. Thank you

PR #377, opened 2018-03-27
detect unknown fields in Kube-manifest.yaml

mengqiy:
This PR depends https://github.com/kinflate/tuthello/pull/2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.9.x
+  - 1.10.x
 
 go_import_path: k8s.io/kubectl
 
--- a/cmd/kinflate/README.md
+++ b/cmd/kinflate/README.md
@@ -40,6 +40,8 @@ and your `PATH` contains `$GOPATH/bin`:
 go get k8s.io/kubectl/cmd/kinflate
 ```
 
+Note: golang 1.10 or newer is required.
+
 ## Demos
 
  * [hello world one-pager](demoHelloWorldShort.md)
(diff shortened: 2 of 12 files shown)


mengqiy:
The tutorial test is failing until kinflate/tuthello#2 merged.

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.9.x
+  - 1.10.x
```
Liujingfang1: Why do you need to bump go version?

mengqiy: `DisallowUnknownFields` is introduced in golang 1.10.

Search `encoding/json` in the golang 1.10 [release notes](https://golang.org/doc/go1.10#minor_library_changes), you will see.

cmd/kinflate/README.md
```diff
@@ -13,6 +13,8 @@ go get k8s.io/kubectl/cmd/kinflate
 
 After running the above command, `kinflate` should get installed in your `GOPATH/bin` directory.
 
+Note: `kinflate` requires 1.10 or newer
```
Liujingfang1: Assume 1.10 is available. When running `kinflate inflate -f . | kubectl apply -f -`, kubectl is also using 1.10. Is that right?

mengqiy: kubectl doesn't have to use golang 1.10. It's just a binary, which can be downloaded or compiled. 
Using golang 1.10 to compile should work fine, it should be backward compatible.

mengqiy:
travis ci didn't report test status...

PR #383, opened 2018-03-28
Return rest config and regular config  from pluginutil Init method

jberkhahn:
-add namespace overriding based on kubectl plugin env var

We added namespace overriding in svcat's standalone mode, so we need to be able to override it in plugin mode as well. I figured while I was in here, I might as well make the init method return the whole config, in case we need to pull anything off it in the future.

Thanks!


jberkhahn:
Anyone? We're blocked on this over on service catalog.

PR #388, opened 2018-04-02
Add glossary

monopole:
and pointers to it from the main README.

Need a place to define terminology.

This is a conversion and update of the content at go/kinflate-glossary

At the last minute, i pulled workflows out of the glossary, to make it a freestanding MD file, to which we can figures.  Perhaps this stuff goes into the main README, but trying to keep that short.

### DIFF
--- a/cmd/kinflate/glossary.md
+++ b/cmd/kinflate/glossary.md
@@ -1,0 +1,324 @@
+# Glossary
+
+[Resource]: #resource
+[application]: #application
+[base]: #base
+[bases]: #base
+[bespoke]: #bespoke-configuration
+[kinflate]: #kinflate
+[manifest]: #manifest
+[overlay]: #overlay
+[overlays]: #overlay
+[off-the-shelf]: #off-the-shelf
+[OTS]: #off-the-shelf
+[patch]: #patch
+[patches]: #patch
+[proposal]: https://github.com/kubernetes/community/pull/1629
+[resource]: #resource
+[resources]: #resource
+[target]: #target
+[TypeMeta]: https://github.com/kubernetes/kubernetes/blob/master/pkg/api/unversioned/types.go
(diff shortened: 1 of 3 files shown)


cmd/kinflate/glossary.md
```diff
@@ -0,0 +1,267 @@
...
+[resource]: #resource
+[resources]: #resource
+[rpm]: https://en.wikipedia.org/wiki/Rpm_(software)
+[target]: #target
+[workflow]: workflows.md
+
+## application
+
+An _application_ is a group of k8s resources that
+server some common purpose, e.g.  a webserver backed by
```
droot: s/server/serves

cmd/kinflate/glossary.md
```diff
@@ -0,0 +1,267 @@
...
+application resource is just another resource.  There's
+some conceptual overlap between the proposal’s
+application resource and the kinflate [manifest].  The
+application resource has a `Components` field that
+serves a purpose similar to the `resources` field in
+the kinflate manifest.  This overlap can be resolved in
+various ways, the simplest being that kinflate does
+nothing special with the application resource.
+
+
```
droot: hmm... not sure if should include "application" here. Do we refer to it anywhere ? Users might think that `kinflate` does something with `application` and the fact it does not.

cmd/kinflate/glossary.md
monopole: done

cmd/kinflate/glossary.md
monopole: * Brian's doc is still called declarative application management (not declarative configuration mgmt), and we're an impl, so we kinda have to address the "a" word

* we've been asked to describe how kinflate works with the proposed application resource, both by folks here and folks in the app-def WG.  Trying to answer that by saying that at the moment it does nothing special.

droot: Sounds good.

PR #389, opened 2018-04-02
add workflow figs

monopole:


cmd/kinflate/README.md
```diff
@@ -1,8 +1,8 @@
 # kinflate
 
-[_kubectl apply_]: glossary.md#apply
-[DAM]: glossary.md#declarative-application-management
-[workflows]: workflows.md
+[_kubectl apply_]: docs/glossary.md#apply
+[DAM]: docs/glossary.md#declarative-application-management
+[workflows]: docs/workflows.md
```
droot: +1 on moving these under docs dir.

cmd/kinflate/docs/workflows.md
```diff
@@ -0,0 +1,164 @@
...
+
+## No local files
+
+One just installs some [OTS] configuration from the web,
+[applying] it directly to the cluster.
+
+This assumes a configuration specified at https://github.com/kinflate/ldap.
+
+> ```
+> kinflate inflate -f https://github.com/kinflate/ldap |\
```
droot: We don't support manifest off the HTTP URL yet, so we should take this out ?

cmd/kinflate/docs/workflows.md
```diff
@@ -0,0 +1,164 @@
...
+> ```
+> kinflate inflate -f ~/myldap | kubectl apply -f -
+> ```
+
+The myldap dir contains a `Kube-manifest.yaml`
+referencing the base via URL, and a `deployment.yaml`
+resource [patch] that increases the replica count
+specified in the base.
+
+## Bespoke configuration
```
droot: we should probably move this to the top ?

cmd/kinflate/docs/workflows.md
monopole: It's so easy to add though 😃

mengqiy:
The figures looks great!
BTW don't forget to squash the commits before merging. They have the same name.

PR #390, opened 2018-04-03
kinflate: added version command

droot:
add version package. Will be using it in container builder to embed version in all the releases. It will help us during debugging any external issues.

### DIFF
--- a/cmd/kinflate/version/version.go
+++ b/cmd/kinflate/version/version.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


cmd/kinflate/version/version.go
```diff
@@ -0,0 +1,67 @@
...
+	return Version{
+		kinflateVersion,
+		gitCommit,
+		buildDate,
+		goos,
+		goarch,
+	}
+}
+
+func (v Version) Print() {
```
mengqiy: I'd like to see it prints to an `io.writer`.
IOW, `NewCmdVersion` takes another argument `stdOut io.writer`.

droot: Done

droot:
@mengqiy Thanks for the quick review. Addressed the comment about the writer.
Example output

```
Version: {KinflateVersion:1.6 GitCommit:$Format:%H$ BuildDate:1970-01-01T00:00:00Z GoOs:unknown GoArch:unknown}
```

In our releases, all the fields will be filled in properly.

PR #404, opened 2018-04-04
Remove the -f option of inflate

monopole:
The manifest path is now just the first arg after the command.
If no arg specified, the current working directory is assumed.

### DIFF
--- a/pkg/kinflate/commands/inflate.go
+++ b/pkg/kinflate/commands/inflate.go
@@ -24,6 +24,8 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"errors"
+
 	"k8s.io/kubectl/pkg/kinflate/app"
 	kutil "k8s.io/kubectl/pkg/kinflate/util"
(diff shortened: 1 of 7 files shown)


pkg/kinflate/commands/inflate.go
```diff
@@ -39,19 +41,14 @@ func newCmdInflate(out, errOut io.Writer, fs fs.FileSystem) *cobra.Command {
...
 
 	cmd := &cobra.Command{
-		Use:   "inflate -f [path]",
+		Use:   "inflate [path]",
 		Short: "Use a Manifest file to generate a set of api resources",
 		Long:  "Use a Manifest file to generate a set of api resources",
 		Example: `
 		# Use the Kube-manifest.yaml file under somedir/ to generate a set of api resources.
-		inflate -f somedir/`,
+		inflate somedir/`,
```
droot: we should remove that extra slash from `somedir/`

PR #408, opened 2018-04-06
Use k8s GCR vanity URL

thockin:


thockin:
I can't tell if this is auto-generated or not?

thockin:
@pwittrock ?

PR #416, opened 2018-04-09
structure for manifest documentation

monopole:
We need some documentation for the manifest on the website.

Here's an initial structure, that uses the "kinflate init" command as a source of truth
to create an example that displays on the website.

To update the docs, create a branch, run "make docs", and merge the branch.

Down the road we could use the makefile to do more sophisticated inclusion of generated code with the website.

### DIFF
--- a/cmd/kinflate/Makefile
+++ b/cmd/kinflate/Makefile
@@ -1,0 +1,25 @@
+SHELL := /bin/bash -euo pipefail
+
+config_file_name = Kube-manifest.yaml
+example_config = docs/$(config_file_name)
+
+.PHONY: all
+all: docs
+
+# In a branch, run 'make docs' to update docs with
+# generated code, then merge it to master.
+docs: $(example_config)
+
+# Use kinflate to create the standard kinflate configuration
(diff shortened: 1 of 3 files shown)


cmd/kinflate/Makefile
```diff
@@ -0,0 +1,28 @@
...
+	echo "# To generate:  make docs" >> TMP
+	echo "################################" >> TMP
+	echo " " >> TMP
+	/tmp/bin/kinflate init
+	cat $(config_file_name) >> TMP
+	mv TMP $(example_config)
+	rm $(config_file_name)
+
+/tmp/bin/kinflate:
+	GOPATH=/tmp go get k8s.io/kubectl/cmd/kinflate
```
droot:
```
go build -o /tmp/bin/kinflate kinflate.go
```
Using above command to build temporary binary will be safer here because it works with current checked out repo as is. If developer has checked out the repo in path other than `k8s.io/kubectl`, current approach will fetch the repo in `k8s.io...` instead of using current repo.

Liujingfang1:
Looks good to me.

PR #428, opened 2018-04-11
change kinflate to kustomize

droot:
Rebased and resolved conflict in `pkg/kustomize/transformers/overlay.go`


mengqiy:
Travis is not happy.
It seems we still use `kinflate` somewhere.

PR #440, opened 2018-04-13
Example kustomize.yaml

### LINKED ISSUE #430: AAUI need to see a completely documented kustomize.yaml example on website (3)

monopole:
Fixes #430

### DIFF
--- a/cmd/kustomize/docs/kustomize.yaml
+++ b/cmd/kustomize/docs/kustomize.yaml
@@ -1,15 +1,101 @@
-namePrefix: some-prefix
-# Labels to add to all objects and selectors.
-# These labels would also be used to form the selector for apply --prune
-# Named differently than “labels” to avoid confusion with metadata for this object
-labelsToAdd:
-  app: helloworld
-annotationsToAdd:
-  note: This is an example annotation
-resources: []
-#- service.yaml
-#- ../some-dir/
-# There could also be configmaps in Base, which would make these overlays
-configMapGenerator: []
-# There could be secrets in Base, if just using a fork/rebase workflow
-secretGenerator: []
+# ----------------------------------------------------
+# Example kustomize.yaml content.
+#
+# This file declares the customization provided by
+# the kustomize program.
+#
+# Since customization is, by definition, _custom_,
+# there are no sensible default values for the fields
+# in this file.
+#
+# The field values used below are merely examples, not
+# to be copied literally.  The values won't work if
+# they happen to be references to external files that
+# don't exist.
+#
+# In practice, fields with no value should simply be
+# omitted from kustomize.yaml to reduce the content
+# visible in configuration reviews.
+# ----------------------------------------------------
+
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+namePrefix: alices-
+
+# Labels to add to all resources and selectors.
+commonLabels:
+  someName: someValue
+  owner: alice
+  app: bingo
+
+# Annotations (non-identifying metadata)
+# to add to all resources.  Like labels,
+# these are key value pairs.
+commonAnnotations:
+  oncallPager: 800-555-1212
+
+# Each entry in this list must resolve to an existing
+# resource definition in YAML.  These are the resource
+# files that kustomize reads, modifies and emits as a
+# YAML string, with resources separated by document
+# markers ("---").
+resources:
+- some-service.yaml
+- ../some-dir/some-deployment.yaml
+
+# Each entry in this list results in the creation of
+# one ConfigMap resource (it's a generator of n maps).
+# The example below creates a ConfigMap with the
+# names and contents of the given files.
+configMapGenerator:
+- name: myJavaServerProps
+  files:
+  - application.properties
+  - more.properties
+
+# Each entry in this list results in the creation of
+# one Secret resource (it's a generator of n secrets).
+# A command can do anything to get a secret,
+# e.g. prompt the user directly, start a webserver to
+# initate an oauth dance, etc.
+secretGenerator:
+- name: app-tls
+  commands:
+    tls.crt: "cat secret/tls.cert"
+    tls.key: "cat secret/tls.key"
+  type: "kubernetes.io/tls"
+
+# Each entry in this list should resolve to a directory
+# containing a kustomization file, else the
+# customization fails.
+#
+# The presence of this field means this file (the file
+# you a reading) is an _overlay_ that further
+# customizes information coming from these _bases_.
+#
+# Typical use case: a dev, staging and production
+# environment that are mostly identical but differing
+# crucial ways (image tags, a few server arguments,
+# etc. that differ from the common base).
+bases:
+- ../../base
+
+# Each entry in this list should resolve to
+# a partial or complete resource definition file.
+#
+# The names in these (possibly partial) resource files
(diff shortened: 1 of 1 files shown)


cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,17 +1,79 @@
...
+# It survives round-tripping that doesn't
+# respect YAML comments.
+kustomizationName: customization for some set of resources
+
+# Value of this field is prepended to the
+# names of all resources.
+namePrefix: foo
+
+# Labels to add to all resources and selectors.
 objectLabels:
```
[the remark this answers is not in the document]
monopole: Yes, as dicussed.

Must do in a different PR that also changes the Go struct and all the examples.

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,17 +1,79 @@
...
-  app: helloworld
+  someName: someValue
+  app: hello
+  foo: bar
+
+# Annotations (non-identifying metadata)
+# to add to all resources.  Like labels,
+# these are key value pairs.
 objectAnnotations:
```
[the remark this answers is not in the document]
monopole: ditto.

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,17 +1,79 @@
...
-description: helloworld does useful stuff.
-namePrefix: some-prefix
-# Labels to add to all objects and selectors.
-# These labels would also be used to form the selector for apply --prune
-# Named differently than “labels” to avoid confusion with metadata for this object
+# A description field that serves only as a comment
+# and is not consumed by kustomize.
+# It survives round-tripping that doesn't
+# respect YAML comments.
+kustomizationName: customization for some set of resources
```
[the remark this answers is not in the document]
monopole: dropped

monopole:
Okay, made the field name changes we discussed.

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,15 +1,105 @@
...
+# ----------------------------------------------------
+
+
+# Value of this field is prepended to the
+# names of all resources, e.g. a deployment named
+# "wordpress" becomes "alices-wordpress".
+namePrefix: alices-
+
+# Labels to add to all resources and selectors.
 labelsToAdd:
```
bgrant0607: `labelsToAdd` and `annotationsToAdd` violate API naming conventions. Too imperative.

https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#naming-conventions

`namePrefix` is not `stringToPrependToName`. :-)

commonLabels and commonAnnotations?

justinsb: I've used `additionalX` in the past, which is declarative (?) but I think feels imperative.

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,15 +1,105 @@
...
+
+# Each entry in this list results in the creation of
+# one Secret resource (it's a generator of n secrets).
+# A command can do anything to get a secret,
+# e.g. prompt the user directly, start a webserver to
+# initate an oauth dance, etc.
+secretGenerator:
+- name: app-tls
+  commands:
+    tls.crt: "cat secret/tls.cert"
```
justinsb: I'm not clear what these commands are.  Are they shell commands?

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,15 +1,105 @@
...
+#
+# The customization declarations in these _base_ files
+# are applied first, in order, to be followed by the
+# customizations declared in this file (the _overlay_).
+#
+# Typical use case: a dev, staging and production
+# environment that are mostly identical but differing
+# crucial ways (image tags, a few server arguments,
+# etc. that differ from the common base).
+bases:
```
bgrant0607: What's the difference between bases and resources?

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,15 +1,105 @@
...
+- ../../base
```
justinsb: So obviously nerd-me likes the idea of a list, but UX-me wonders if it's overkill.

As an example:

```
bases:
- ../1
- ../2
```

Does everything in `2` need to reference things in `1`? (if so, I'd think we'd simply have `1` as the base for `2`)

Or are we combining two things that are effectively two packages - e.g. 1 is `apache-wordpress` and 2 is `mysql` ?

cmd/kustomize/docs/kustomize.yaml
monopole: Thanks for the feedback!

`commonLabels`, `additionalLabels` are nouns with adjectives.  `labelsToAdd` is a noun with a prepositional phrase functioning as an adjective.  `addLabels` would certainly be imperative.  

It would be nice to go with simply `labels` and `annotations`, to keep it consistent with (as Brian says) `namePrefix` and `resources`.

But shouldn't kustomization support _removal_?  If so, we want a struct that imposes no syntactic cost on people that have no need for removal (since we think removal would be rare).

Field names like `fooToAdd` and `fooToRemove` do this nicely - just omit `fooToRemove`.  Plus the names are self-documenting.

Examples of syntax that impose more cost on people with no need for removal are:
```
stuffToRemove {
  labels:
    fruit : apple
  resources:
  - map1.yaml
}
stuffToAdd {
  labels:  
    fruit : peach
  resources:
  - dep.yaml
  - map2.yaml
}
```
or the horrible
```
labels:
- operation: remove
  map:
    fruit: apple
- operation: add
  map:
    fruit: peach
resources:
- operation: remove
  - map1.yaml
- operation: add
  - dep.yaml
  - map2.yaml
}
```

cmd/kustomize/docs/kustomize.yaml
monopole: Brian wanted a list :)
It does seem a little too powerful and confusing.

cmd/kustomize/docs/kustomize.yaml
```diff
@@ -1,15 +1,105 @@
...
+ - ../some-dir/some-deployment.yaml
+
+# Each entry in this list results in the creation of
+# one ConfigMap resource (it's a generator of n maps).
+# The example below creates a ConfigMap with the
+# names and contents of the given files.
+configMapGenerator:
+- name: myJavaServerProps
+  files:
+    - application.properties
```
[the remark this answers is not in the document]
monopole: yes, it creates a map named `myJavaServerProps` with two keys (file names), and two values (file contents).   You want to provide a means to override the value of the key?

monopole:
We could also just not support delete, and explain why.

cmd/kustomize/docs/kustomize.yaml
monopole: Simpler to not support delete, and say way.
 * don't add it to you base.
 * if you don't control the base, fork the base and remove it there.

cmd/kustomize/docs/kustomize.yaml
monopole: BTW, it's the latter.  Everything from these various bases, regardless of file system hierarchy, gets sucked into the non-hierarchical sea of resources (module name collisions, which is a detected error).

monopole:
updated to commonLabels, commonAnnotations

PR #460, opened 2018-05-01
drop unused pngs

monopole:
Fix some nits (bad links, kustomize in normal font, etc.) and drop two unused png files.


cmd/kustomize/README.md
```diff
@@ -4,6 +4,7 @@
 [base]: docs/glossary.md#base
 [declarative configuration]: docs/glossary.md#declarative-application-management
 [demo]: demos/README.md
+[demos]: demos/README.md
```
Liujingfang1: demos and demo point to the same file. Do you want to keep both of them?

monopole: yes.  only way to do plural

monopole: it's a markdown thing.  Want to be able to use \[demo\] and \[demos\] in text as links.

PR #516, opened 2018-07-31
[WIP] Kubectl main out of core repo with vendored dependencies

seans3:
* To build `kubectl` binary (top-level directory): `go build kubectl.go`
* Pulls only one file (`cmd/kubectl/kubectl.go`, `kubectl` main) out of the core repo.
* Doesn't move any other `kubectl` code.
* Vendors all remaining `kubectl` dependencies.
* Allows decomposition of `kubectl`, including some `kubectl` subcommands in different repos.
* Allows for multiple ownership of pieces of `kubectl`.
* Allows more frequent releases, and different release cadences.
* Simplifies testing.


apelisse:
Could you please describe the steps that lead you here? It's just impossible to know what's going on here.

Can you move the `kubectl.go` file in a separate commit?

deads2k:
This goes directly against the clear guidance from @kubernetes/sig-architecture-pr-reviews that says we do not vendor k8s.io/kubernetes from other repos.  If you want to have `kubectl` as a separate repo, the proper path to do that is to continue eliminating the dependencies on internal types and once they are gone it will be possible to build as a separate repo.

Vendoring k8s.io/kubernetes is a bad a idea and is likely to lead to significant drift and/or an inability to reasonably revendor and keep up. We have a lot of experience doing this.  If you want to move into a separate repo, first snip the links.

@smarterclayton @liggitt @thockin we've all talked about this before.

pwittrock:
> This goes directly against the clear guidance from @kubernetes/sig-architecture-pr-reviews that says we do not vendor k8s.io/kubernetes from other repos.

One of the big motivation for this is that the staging work done by apimachinery breaks vendoring dependencies built in other repos back in to kubernetes/kubernetes and undermines efforts to break components out in a modular or incremental manner.  Are there any plans to resolve this?

> Vendoring k8s.io/kubernetes is a bad a idea and is likely to lead to significant drift and/or an inability to reasonably revendor and keep up. We have a lot of experience doing this. If you want to move into a separate repo, first snip the links.

@deads2k You suggested this 8 months ago and said it should be possible in a quarter, to which I responded that my estimation is that it was much more work, but would happily give you the opportunity to snip the dependencies yourself.  Have you made any progress here?

pwittrock:
@deads2k After talking offline seems like you made a lot more progress than I expected.  Do you have an estimation for how much investment is still required?

soltysh:
This was discussed during last SIG-CLI call. We have a somewhat clear path towards extracting the repo. With that in mind I'll go ahead and close this PR.

Issue #588, opened 2019-02-11
Allow kubeconfig to be Base64 encoded string


webdeb:
**FEATURE REQUEST**

kubectl --kubeconfig=$K8S_CONFIG_BASE64 [do something]

Would be nice, if you could just pass the whole kubeconfig as base64 string to kubectl command.
The main reason for this is to use kubectl in CI/CD pipelines, where people are storing their kubeconfig in some sort of settings/variables (like gitlabs secrets). Also if you use some kind of kubectl container to run kubectl.

AFAIK the mainly used workaround is to decode that string, create a tmp/file, and then executing the command.

webdeb:
Just an idea that could be implemented very easily: 

```
--kubeconfig="base64:$K8S_CONFIG"
```

waldbeerlein:
+1

webdeb:
@seans3 please check out my PR

mishushakov:
You could just:

1. Copy your kubeconfig to clipboard as base64 string:
`cat ~/.kube/config | base64 | pbcopy `

2. On your CI pipeline, create a environment variable `kube_config` and paste your base64 string
3. Encode your base64 variable back to kubeconfig environmental variable:
`echo ${kube_config} | base64 -d > ${KUBECONFIG}`

4. Set context with kubectl:
`kubectl config use-context <name of your context>`
5. Execute your commands
6. Done!

webdeb:
@mishushakov yes, thats exactly what I already wrote. these are 4 commands you have to put inside your ci/cd scripts, and **unfortunately this method does not work if you store the variable in gitlab secrets** because it prevents secrets to be printed.

mishushakov:
@webdeb on GitLab you can connect your cluster and have access to k8s environment variables (https://gitlab.com/help/user/project/clusters/index#deployment-variables), in this case we actually need the `KUBECONFIG` variable:

<img width="1230" alt="Screenshot 2019-04-05 at 02 05 26" src="https://user-images.githubusercontent.com/10400064/55596102-4a28e780-5747-11e9-9b0e-70b905eb983b.png">

cbluth:
I have been looking for something like this, to execute `kubectl` commands against different clusters progamatically without having to deal with KUBECONFIG files on the filesystem.

unixfox:
Can someone reopen this issue? Some people are still interested into this feature.

webdeb:
Just a hint for everyone who needs this for the GitLab CI.

[...]

PR #590, opened 2019-02-15
Feature: allow to pass kubeconfig as a base64 string

webdeb:
This PR makes it possible to provide --kubeconfig="base64:$KUBE_CONFIG_BASE64" as proposed in the issue #588 

---

Hello Guys, this is my first code in go ever, please be cool, if something is missing, docs, tests, code, whatever.. no big deal, I'll fix it.

thx

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,182 +2,450 @@
 
 
 [[projects]]
+  digest = "1:8e47871087b94913898333f37af26732faaab30cdb41571136cf7aec9921dae7"
   name = "github.com/PuerkitoBio/purell"
   packages = ["."]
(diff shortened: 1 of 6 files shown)


liggitt:
Thanks for the pull request. Code for kubectl and client-go lives in Kubernetes/Kubernetes, not in this repo. 

Also, since this affects client-go, not just kubectl, overloading the current flag would cause all existing clients/libraries to fail to parse that flag consistently. 

The proposed feature has broad enough impact that it should start as a KEP and be accepted before implementing. See https://github.com/kubernetes/enhancements/blob/master/keps/0001-kubernetes-enhancement-proposal-process.md for details

/close

PR #624, opened 2019-03-26
Remove cp examples from the port-forward page

jonas:
The examples at the end of the port_forward_to_pods page are derived
from the copying_container_files page when it was created.


DanyC97:
@jonas any chance you can update the Pr description with the rational on why you think is needed to remove the content ?

Rio:
It's a left over from `copying_container_files.html`. That file clearly got copied and the last 2 examples didn't get removed.

jonas:
@DanyC97 I will update the commit message. As @Rio pointed out, it looks like those examples came from another page.

jonas:
Amended the commit message.

jonas:
Fixed conflict #635

PR #634, opened 2019-03-28
feat(docs): add section on environment style configmap/secret generation

### LINKED ISSUE #632: Configmap/Secret generation documentation is missing an option for generating from environment files
In the section `Secrets and ConfigMaps` there are explanations about how to generate configmaps using the `literals` and `files` options. One is missing however which is the `env` option that generates the same type of configmaps as the `literals` option but sources the key/value pairs from an environment file.

Also it's not very clear that the options used on `configMapGenerator` are also used in `secretGenerator`.

Rio:
This adds an example of generating configmaps/secrets using the `env` option.
It also adds a bit of clarification that the same `literals`, `files` and `env` options can be used for secrets.

closes #632

### DIFF
--- a/docs/book/pages/app_management/secrets_and_configmaps.md
+++ b/docs/book/pages/app_management/secrets_and_configmaps.md
@@ -112,6 +112,56 @@ data:
 ```
 {% endmethod %}
 
+### ConfigMaps From Environment Files
+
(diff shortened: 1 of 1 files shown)


pwittrock:
@Rio Thanks!

/approve
/lgtm

PR #653, opened 2019-05-10
migrate to go modules

tariq1890:
Why do we need this PR?

i) The vendor directory is terribly out-of-date. I have updated the vendor directory first using `dep ensure` and then ran `go mod init`

ii) Keeping up with the times.


seans3:
Please note that this repository is going to be cleared in the near future in preparation for accepting kubectl code from Kubernetes staging. The code published into this repository from Kubernetes staging will already have go module support, so this PR is unnecessary.

This repository has been backed up. We will keep the issues. We will re-introduce the parts we care about into Kubernetes staging. 

Please contact me if you have any questions or concerns.

Issue #910, opened 2020-08-14
Ensure we're using i18n where possible


eddiezane:
There is a scattered usage of `i18n.T()` throughout descriptions and flags. Let's make future translations easier by ensuring we're using i18n where feasible.

Examples:

https://github.com/kubernetes/kubectl/blob/19b50d838ed2e2a7389a3fb467b259982c772b73/pkg/cmd/apply/apply.go#L171
https://github.com/kubernetes/kubectl/blob/19b50d838ed2e2a7389a3fb467b259982c772b73/pkg/cmd/annotate/annotate.go#L153

/sig cli
/kind cleanup
/priority backlog
/good-first-issue

Evalle:
I think I can help with that. 
/assign

jayendra13:
What kind of pattern should I follow to internationalise the strings, is it applicable to only the help text for commands ?

eddiezane:
@jayendra13 let's aim to get command and flag descriptions wrapped and then evaluate what else you find.

jayendra13:
> @jayendra13 let's aim to get command and flag descriptions wrapped and then evaluate what else you find.

👍 got it.

HerrNaN:
@jayendra13 @eddiezane It seems like nothing has happened here for a while, is it ok if I take over?

jayendra13:
yes sure I got busy with some other stuff, and I almost forgot this.
you can continue on this.

jayendra13:
may be this can help https://github.com/kubernetes/kubectl/pull/916

HerrNaN:
Somehow adding i18n translations on every single `Short` and `Long` line seems a bit micromanaging in a way. Wouldn't it be easier to go through all the commands and add that separately? This would allow the code that describe how the commands should work less cluttered, more consise and easier to read and we can ensure that all commands get these translations.

WDYT @eddiezane?

rikatz:
xref: https://github.com/kubernetes/kubernetes/pull/97610

PR #916, opened 2020-08-19
Using i18n for command helps

jayendra13:
Fixing https://github.com/kubernetes/kubectl/issues/910

### DIFF
--- a/pkg/cmd/annotate/annotate.go
+++ b/pkg/cmd/annotate/annotate.go
@@ -131,7 +131,7 @@ func NewCmdAnnotate(parent string, f cmdutil.Factory, ioStreams genericclioption
 		Use:                   "annotate [--overwrite] (-f FILENAME | TYPE NAME) KEY_1=VAL_1 ... KEY_N=VAL_N [--resource-version=version]",
 		DisableFlagsInUseLine: true,
 		Short:                 i18n.T("Update the annotations on a resource"),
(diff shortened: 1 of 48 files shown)


eddiezane:
@jayendra13 thanks for taking this one one!

The repo setup is a bit confusing at the moment and this repo is only used for tracking kubectl issues. All PR's need to be made against the [main kubernetes repo](https://github.com/kubernetes/kubernetes). kubectl lives under [staging/src/k8s.io/kubectl](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kubectl) there. That code is then copied here as a periodic job. This repo will eventually be the source of truth which is why it currently lives under staging while we migrate it out.

Would you kindly close this PR and reopen it there?

https://github.com/kubernetes/kubectl/blob/master/CONTRIBUTING.md

brianpursley:
+1 to what Eddie said

But also, have you guys seen this?  https://github.com/kubernetes/kubernetes/blob/2c2ca27bfc538733c0f0ddbab7d3f2ff8707403d/translations/extract.py

Are we supposed to be running that python script?

I’m not too familiar with how kubernetes i18n works, but I wonder how broken it is because I don’t think people have been keeping it up. For example, if there is a long multi line description of a command, and one little thing changes , does that break the translations?

jayendra13:
Ok will do that,
I will also check the k8s implementation of i18n works in the similar way.

PR #918, opened 2020-08-19
Correct a small typo in the Images section

mattKaczorowski:
Sorry, we do not accept changes directly against this repository. Please see
CONTRIBUTING.md for information on where and how to contribute instead.

### DIFF
--- a/docs/book/pages/reference/kustomize.md
+++ b/docs/book/pages/reference/kustomize.md
@@ -300,7 +300,7 @@ Definitions:
 | Name          | Type    | Desc                                |
 | :------------ | :------ | :---------------------------------- |
 | **name** | string  | Match all `image` fields with this value for the *name*|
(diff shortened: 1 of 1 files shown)


eddiezane:
@mattKaczorowski thanks for catching and fixing this!

The repo setup is a bit confusing at the moment and this repo is only used for tracking kubectl issues. All PR's need to be made against the [main kubernetes repo](https://github.com/kubernetes/kubernetes). kubectl lives under [staging/src/k8s.io/kubectl](https://github.com/kubernetes/kubernetes/tree/master/staging/src/k8s.io/kubectl) there. That code is then copied here as a periodic job. This repo will eventually be the source of truth which is why it currently lives under staging while we migrate it out.

Would you kindly close this PR and reopen it there?

https://github.com/kubernetes/kubectl/blob/master/CONTRIBUTING.md

PR #983, opened 2020-11-25
Show the available container names when the user misses the container name in exec

dramasamy:
Show the available container names when the user misses -c/--container option in the 'kubectl exec' command for the multi-container pod; this saves time and improves the user experience instead of executing the 'pod describe' command.


dramasamy:
I'll do this in 'staging'

PR #1143, opened 2021-11-08
Print creation time when viewing the details of a rollback version info。

lijinfeng2011:
Before performing the rollback operation, you need to obtain the creation time of the corresponding version to determine whether to roll back this version.

The instructions involved are: kubectl rollout history TYPE NAME -- revision=ID

------
When option “--revision” is used, “-o yaml” does not take effect.
So far, there is no a way to get the revision's created time. This modification can temporarily solve this problem.

### DIFF
--- a/pkg/polymorphichelpers/history.go
+++ b/pkg/polymorphichelpers/history.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"time"
 	"text/tabwriter"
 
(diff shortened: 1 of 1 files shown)


lauchokyip:
Hi @lijinfeng2011 , please open your PR to the main repo k/k repo https://github.com/kubernetes/kubernetes. 

kubectl is in the process of moving out the main repo so in the future you can open a PR here in the futrue. Thank You.

Issue #1455, opened 2023-07-29
Default `kubectl autoscale` to `ApiVersion=autoscale/v2`


beejiujitsu:
**What would you like to be added**:

Change the default `kubectl autoscale` result from using `apiVersion: autoscale/v1` to `apiVersion: autoscale/v2`.

https://github.com/kubernetes/kubectl/blob/197123726db24c61aa0f78d1f0ba6e91a2ec2f35/pkg/cmd/autoscale/autoscale.go#L263

**Why is this needed**:

Generating v2 would mean the [HPA documentation examples, specs, etc.](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale), would reflect proper use/modification with resources generated with `kubectl autoscale`. 

While it requires a newer k8s version, v1.23+, that was released nearly 2 years ago; it's time to update kubectl.

PR #1456, opened 2023-08-01
Updated pkg/cmd/autoscale/autoscale.go

akash-kumar-saw:
Issue Number #1455,

Changed the default kubectl autoscale from using apiVersion: autoscale/v1 to apiVersion: autoscale/v2.


ardaguclu:
Thanks for opening PR. However, this repository does not accept any direct pull requests. Instead PRs should be opened in kubernetes repository and it is mirrored automatically in here.
/close

PR #1678, opened 2024-11-13
refactor(create job): remove redundant nil check

miyamo2:
Removed the nil check for a slice that was performed before it's length check.
This is linted in staticcheck as well.
https://staticcheck.dev/docs/checks#S1009


ardaguclu:
We don't get pull requests directly in this repository. Instead it is automatically mirrored from kubernetes.
/close

PR #1719, opened 2025-02-20
describeSecret: sort alphabetically to avoid random order

HaraldNordgren:


HaraldNordgren:
@ardaguclu Sorry, I noticed now and opened the same change here instead: https://github.com/kubernetes/kubernetes/pull/130309

PR #1794, opened 2025-10-16
no need to cost time for function call

xiyanxiyan10:
no need to cost time for function call


ardaguclu:
PRs should be opened in main repo, not in here
/close

